PR #3, opened 2015-01-17
Classes for protein/transcript variant effects

iskandr:
- changed `mutate` to use `transcript_effects` objects instead of single `ProteinMutation` object


iskandr:
Merging after a code walkthrough with @ihodes and @tavinathanson.

PR #7, opened 2015-02-10
Refactor core logic

iskandr:


varcode/common.py
```diff
@@ -59,3 +59,14 @@ def trim_shared_flanking_strings(ref, alt):
     ref, alt, prefix = trim_shared_prefix(ref, alt)
     ref, alt, suffix = trim_shared_suffix(ref, alt)
     return ref, alt, prefix, suffix
+
+
+def group_by(records, field_name):
```
ihodes: I'd imagine you're doing this instead of using itertools.group_by because that one returns generators instead? As an aside, this is a great place to use a collections.defaultdictionary.

iskandr: 1) Yep, I didn't want the small annoyance of converting the iterator to a dictionary at the call site. 

2) Good point about `defaultdict`!

varcode/transcript_mutation_effects.py
```diff
@@ -313,7 +315,7 @@ def get_class(effect):
             return _MultipleSubstitution
     return effect.__class__
 
-variant_effect_priority_list = [
+transcript_effect_priority_list = [
```
timodonnell: this get_class function and the fake _MultipleSubstitution class seem hacky to me. Could this logic and also the places where the top ranked variant is selected be instead done with a sort_variant_effect_by_priority(variants) function? All of this logic could then at least be internal to that function. Or, maybe better: just make a function called variant_effect_priority(variant) -> int that returns a priority number for that variant, and can then be used as a key function in sorted()

timodonnell: if you go the latter route, I'd suggest implementing that function in the parent class (TranscriptMutationEffect), so all variant effects have a .priority (int) property.

varcode/variant_collection.py
```diff
@@ -127,8 +126,27 @@ def __str__(self):
...
-        ]
\ No newline at end of file
+            self.annot.describe_variant(variant)
+            for variant
+            in self.records
+        ]
+
+    def print_variant_effects(self):
+        for variant_effect in self.variant_effects():
+            transcript_effect_count = 0
```
timodonnell: I can make a PR later that makes this code python 3 friendly

varcode/common.py
```diff
@@ -59,3 +60,12 @@ def trim_shared_flanking_strings(ref, alt):
     ref, alt, prefix = trim_shared_prefix(ref, alt)
     ref, alt, suffix = trim_shared_suffix(ref, alt)
     return ref, alt, prefix, suffix
+
```
timodonnell: docstring?

varcode/core_logic.py
```diff
@@ -367,12 +357,14 @@ def infer_transcript_effect(variant, transcript):
             "Expected %s : %s to have type Transcript" % (
                 transcript, type(transcript)))
 
-    if not transcript.complete:
-        return IncompleteTranscript(variant, transcript)
-
+    # check for non-coding transcripts before whether it's complete since
+    # everyone non-coding transcript is "incomplete".
```
timodonnell: "everyone"

timodonnell:
having never looked at this code before, I'm not sure if my feedback is very helpful, but besides my small nit comments, lgtm

varcode/transcript_mutation_effects.py
iskandr: I agree this is hacky, I wasn't sure how to do it better. I'm a little confused about your first suggestion, how is it different from the `top_priority_transcript_effect` function below the priority list?

As for the second suggestion, that's actually how I started out. However, every time I added a variant I had to go looking around every existing class to figure out what priority number to assign the new class, sometimes having to shift around existing priorities. Seems strictly less maintainable/understandable than having a list of classes.

ihodes: I'd agree re: the having a list of classes being very clear. Variant effects could have a computed property, `.priority`, which deduces from that list what their priority is.

varcode/core_logic.py
iskandr: :+1:

varcode/variant_collection.py

varcode/transcript_mutation_effects.py
iskandr: One possibility to make this a *little* less hacky is to use  `Substitution` to mean single amino acid changes, and use `ComplexSubstitution` or `MultiSubstitution` for multiple amino acid changes. 

The hierarchy would then be:

`Substitution` < `Insertion` < `Deletion` < `MultipleSubstitution`

timodonnell:
LGTM

PR #8, opened 2015-02-13
Add support for Python 3

timodonnell:
Unit tests run for me on both Python 3.4 and 2.7.


test/test_dbnsfp_validation.py
```diff
@@ -65,5 +65,5 @@ def test_dbnsfp_validation_set():
 if __name__ == '__main__':
     for test_tuple in test_dbnsfp_validation_set():
         f = test_tuple[0]
-	args = test_tuple[1:]
-	f(*args)
+    args = test_tuple[1:]
```
iskandr: Why did the actual calls to the test function come out of the loop?

varcode/filter_variants.py
```diff
@@ -31,8 +31,8 @@
     for record in VariantCollection(args.input):
         chrom, pos = record.contig, record.pos
         ref, alt = record.ref, record.alt
-        print chrom, pos, ref, alt
-        print "--", annot.describe_variant(chrom, pos, ref, alt)
+        print(chrom, pos, ref, alt)
```
iskandr: `from __future__ import print_function`?

varcode/infer_effect.py
```diff
@@ -177,7 +177,7 @@ def infer_coding_effect(
     # In case sequence isn't a multiple of 3, then truncate it
     # TODO: if we get a frameshift by the end of a CDS (e.g. in the stop codon)
     # then we should use some of the 3' UTR to finish translation.
-    truncated_variant_cds_seq = variant_cds_seq[:len(variant_cds_seq) / 3 * 3]
+    truncated_variant_cds_seq = variant_cds_seq[:int(len(variant_cds_seq) / 3) * 3]
```
iskandr: These are the cases I'm most worried about in the 2to3 conversion. This one looks correct but be really careful with integer divisions. Once one of those bugs creeps in it'll be very annoying to discover.

varcode/maf.py
```diff
@@ -61,7 +61,7 @@ def load_maf_dataframe(filename, nrows=None, verbose=False):
         filename,
         skiprows=lines_to_skip,
         sep="\t",
-        usecols=range(len(MAF_COLUMN_NAMES)),
+        usecols=list(range(len(MAF_COLUMN_NAMES))),
```
iskandr: I don't know if I'll remember to keep coding like this (constructing every list range twice in Python 2.7.x feels like a wart and something I suspect I'll forget to do).

iskandr:
Other than the indent change in the __main__ section of a test and the lack of `import print_function`, LGTM.

test/test_dbnsfp_validation.py
timodonnell: yikes -- good catch. there was some tab / space issue that i was trying to correct here, but got it wrong. fixed.

varcode/infer_effect.py
timodonnell: yeah, totally agree. i was fairly careful but we should triple check these.

varcode/maf.py
timodonnell: actually, in this particular case the list() was overly cautious and wasn't necessary. removed.

varcode/infer_effect.py
timodonnell: i'm also going to add an "from __future__ import division" so we at least have the same semantics between 2 and 3 for these.

timodonnell:
PR updated to address comments

PR #13, opened 2015-02-17
collect effect annotation errors in dictionary, only look up overlapping...

iskandr:
... transcripts and construct the Gene objects later


varcode/variant_annotator.py
```diff
@@ -11,7 +12,7 @@ class VariantAnnotator(object):
     def __init__(self, ensembl_release):
         self.ensembl = pyensembl.EnsemblRelease(ensembl_release)
 
-    def describe_variant(self, variant):
+    def describe_variant(self, variant, raise_on_error=True):
```
tavinathanson: Should `raise_on_error` be in the `VariantAnnotator` constructor instead of peppered across different methods?

varcode/effect_ordering.py
```diff
@@ -52,10 +52,18 @@ def top_priority_transcript_effect(effects):
...
 
-    def effect_cds_len(effect):
-        """
-        Returns length of coding sequence of transcript associated with effect
-        """
-        return len(effect.transcript.coding_sequence)
-
-    return max(best_effects, key=effect_cds_len)
\ No newline at end of file
+    if any(effect.transcript.complete for effect in best_effects):
```
tavinathanson: I don't really understand the context here, but why not apply the same sorting to both (first by CDS length, then by transcript length), vs. filtering in one case and not in the other?

(Probably a silly question. I'm just not clear on the context.)

varcode/variant_annotator.py
```diff
@@ -40,14 +40,38 @@ def describe_variant(self, variant):
...
+                except (AssertionError, ValueError) as error:
+                    if raise_on_error:
+                        raise
+                    else:
+                        logging.warn(
+                            "Encountered error annotating %s for %s: %s",
+                            variant,
+                            transcript,
+                            error)
+                    errors[transcript] = error
```
tavinathanson: Why not tie each error directly to the transcript effect inference that it came from, vs. having a parallel mapping?

varcode/variant_annotator.py
```diff
@@ -40,14 +40,38 @@ def describe_variant(self, variant):
...
         for (gene_id, transcripts) in overlapping_transcript_groups.items():
             effects = []
             for transcript in transcripts:
-                effect = infer_transcript_effect(variant, transcript)
-                effects.append(effect)
+                try:
+                    effects.append(infer_transcript_effect(variant, transcript))
+                except (AssertionError, ValueError) as error:
+                    if raise_on_error:
+                        raise
```
tavinathanson: No error for the raise?

varcode/variant_annotator.py
```diff
@@ -40,14 +40,38 @@ def describe_variant(self, variant):
...
+                            "Encountered error annotating %s for %s: %s",
+                            variant,
+                            transcript,
+                            error)
+                    errors[transcript] = error
             gene_transcript_effects_groups[gene_id] = effects
 
+        # construct Gene objects for all the genes containing
+        # all the transcripts this variant overlaps
+        overlapping_genes = [
```
tavinathanson: So, in practice, the difference between this result and `self.ensembl.genes_at_locus(contig, pos, end_pos)` is what? I'm a bit confused.

varcode/variant_annotator.py
iskandr: The most common use case is `VariantCollection` which uses `VariantAnnotator` under the hood. Not sure how to economize on `raise_on_error` since you'll need one arg in `VariantCollection` and another in `VariantAnnotator`.

varcode/variant_collection.py
```diff
@@ -82,7 +82,7 @@ def __init__(self, filename, drop_duplicates=True):
         elif filename.endswith(".maf"):
             self.raw_reference_name, self.records = _load_maf(filename)
         else:
-            raise ValueErrr("Unrecognized file type: %s" % (filename,))
```
tavinathanson: I think I prefer `ValueErrr`!

tavinathanson:
A few comments, but generally looks good to me!

varcode/variant_annotator.py
tavinathanson: Yeah, I figured `VariantCollection` would pass it through to `VariantAnnotator`. The constructor just felt more appropriate to me, based on knowing far less about the code than you do.

varcode/effect_ordering.py
iskandr: I'm not sure if I understand the question, so I'm not sure if this clarifies the assumptions: (1) the impact on a coding transcript is more interesting than the impact on a non-coding transcript (2) lacking any other info, a longer CDS is more interesting than a shorter one, (3) non-coding transcripts don't have a CDS, so instead sort them by full transcript length.

varcode/variant_annotator.py
iskandr: I'm not sure what you mean by "the transcript effect inference that it came from", can you say more?

varcode/variant_annotator.py
iskandr: I just thought it was inelegant (and a little wasteful) to do two queries to see what's overlapping the mutated locus. So, instead, I just look up the overlapping transcripts and construct the Gene objects from the transcripts.

varcode/variant_collection.py
iskandr: I agrrree!

varcode/variant_annotator.py
iskandr: What do you mean?

(`raise` by itself simply re-raises the capture exception)

varcode/variant_annotator.py
iskandr: The constructor seems a little clumsier to me since the typical usecase seems to be:

1) Construct `variant_collection = VariantCollection(some_vcf_path)`
2) Call `variant_collection.variant_effects()`
3) Encounter an error
4) Decide the error is irrelevant
5) Call `variant_collection.variant_effects(raise_on_error=False)` to ignore the error

Having to create a new `VariantCollection` seems like a very minor hassle but it also doesn't have a clear payoff.

varcode/variant_annotator.py
tavinathanson: I didn't realize it did that :)

varcode/effect_ordering.py
tavinathanson: I guess the question is: why not leave the non-coding transcript stuff in there (when transcripts do have complete coding sequence annotations), and just have it lower in the sort order vs. filtering it out? Not saying you should. Just making sure I understand the choices.

varcode/variant_annotator.py
tavinathanson: When you error on an `infer_transcript_effect(variant, transcript)`, you collect all the errors into a map from transcript to error. Thoughts on putting the error in the transcript effect object itself, instead of a map from transcript to error in the variant effect object?

varcode/variant_annotator.py
tavinathanson: Ah, so it's the same result

varcode/effect_ordering.py
iskandr: Maybe I'm being unimaginative: how do you get the CDS length of a transcript without a CDS? `np.inf`? Also, is there a case where the result would be different, or are you just saying it would be more elegant to have a single sort key function?

varcode/variant_annotator.py
tavinathanson: :+1:

varcode/variant_annotator.py
iskandr: Any ideas on what these transcript effects would look like? Would I just have a generic `ErrorEffect` object? Maybe once the class of errors is narrowed down to some well understood types of mistakes in the annotation database I can make those into different `Effect` types, for now I'm just catching `ValueError`s and `AssertionError`s from many different sources and don't have any extra information to unify them other than "something went wrong while trying to annotate this variant".

varcode/effect_ordering.py
tavinathanson: Looks like my comment had two parts!

Part of my thinking was the elegance of a single sort key.

But also, in terms of the difference in results: the results for the `if` clause can either contain both complete and incomplete transcript effects, or just complete transcript effects. Just curious about leaving the incomplete stuff in, but lower in the sort order.

Probably not worthy of too much conversationing :)

varcode/variant_annotator.py
iskandr: Yep, just computed via one overlap query instead of two.

varcode/variant_annotator.py
tavinathanson: Aha, it wasn't entirely clear to me that we wouldn't *have* a TranscriptEffect object because we errored in creating it.

Sure, `ErrorEffect` could work. I don't feel strongly, though. Feel free to carry on :)

varcode/effect_ordering.py
iskandr: ...but why leave in incomplete transcripts if we know for sure we're not going to use them? We can't compare their CDS length, so the key would have to be something like:
```
def key_fn(effect): 
   c = effect.transcript.complete
   cds_len = len(effect.transcript.coding_sequence) if c else 0
   return c, cds_len
```

I find that less clear than explicitly dropping the incomplete terms. 

If we wanted to unify both branches we'd need something like:

```
def key_fn(effect): 
   c = effect.transcript.complete
   cds_len = len(effect.transcript.coding_sequence) if c else 0
   return c, cds_len, len(transcript)
```

Which is more elegant but I think less clear on what's happening.

varcode/variant_annotator.py
iskandr: Yeah, it's not really possible to make a e.g. `Substitution` if you raise
an exception while trying to get the original amino acids you're
substituting, the necessary information never gets computed.

varcode/effect_ordering.py
tavinathanson: I just wasn't clear how/when we'd use the incomplete transcripts in the `else` case, and bubbled that lack of clarity to ask about it for the `if` case.

iskandr: All the transcripts in the else case e.g. `Incomplete`, `Noncoding`, or something else without a CDS. If we have at least one coding transcript, we should ignore all those without a CDS. Does that make sense?

(also, this is an axiom/assumption, not necessarily something well grounded in evidence. I could imagine a long ncRNA being the most important transcript of a gene)

tavinathanson: Yup, that logic makes sense to me -- I just wanted to hear more about the assumption it was based on

Issue #16, opened 2015-02-17
empty variant collection when loading strelka vcf


timodonnell:
Loading the below strelka shared strelka VCF, I seem to get an empty variant collection back.

```
>> vc = varcode.VariantCollection("/Users/tim/Box Sync/pt189/variants/strelka-somatic-primary-illumina-exome/passed.somatic.snvs.vcf")
>> len(vc)
0

**Comment by iskandr** [comment]
Huh, when I try to just parse the file with PyVCF I get this exception:

```
/usr/local/Cellar/python/2.7.6/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcf/parser.pyc in next(self)
    529         if self._prepend_chr:
    530             chrom = 'chr' + chrom
--> 531         pos = int(row[1])
    532
    533         if row[2] != '.':
```

iskandr:
The filter column for these VCFs isn't set, changed the parsing logic to allow empty lists as if they were "PASS".

Issue #1, opened 2014-12-03
Annotate essential splice site mutations


hammer:
@tmfromer mentioned that [PLINK/SEQ will do this annotation](https://bitbucket.org/statgen/plinkseq/src/master/sources/plinkseq/sources/lib/annot.cpp?at=master#cl-846).

iskandr:
Thanks for this link, I'll try to keep the mutation annotations and their relative priorities similar to PLINK/SEQ.

iskandr:
Useful reference: [Intronic and Exonic Nucleotide Variations that Affect RNA Splicing in Humans](https://www.iconceptpress.com/download/paper/12051922401407.pdf)

>The consensus sequence of the 5’ splice site is “CAG/GUAAGUAU”, where “/” indicates an exon-intron boundary

>The 3’ splice site consists of 3’ end of an intron and 5’ end of an exon, having a consensus sequence of “NYAG/R” (N, any nucleotide; Y, pyrimidine; R, purine; “/”, intron-exon boundary)

>It has been predicted that 16-20% of missense mutations disrupts ESEs and change the splicing pattern of the gene (Gorlov et al., 2003). Comparison of effects of mutations in HGMD and neutral SNPs in dbSNP on ESE/ESS hexamers and subsequent validation of an example of de novo generated ESSs revealed that 7,154 out of 27,681 missense and nonsense mutations (26%) are likely to cause exon skipping (Sterne- Weiler et al., 2011).

iskandr:
Still missing: splice site mutations on the exon side.

PR #17, opened 2015-02-18
added IntronicSpliceSite, SpliceDonor, SpliceAcceptor effects

iskandr:


iskandr:
Fixes https://github.com/hammerlab/varcode/issues/16 and partially fixes https://github.com/hammerlab/varcode/issues/1 (still need to support exonic splice variants)

varcode/core_logic.py
```diff
@@ -381,6 +383,78 @@ def infer_exonic_effect(variant, transcript):
...
+            assert exon.start <= end and exon.end >= start, \
+                "Expected interval %d:%d to overlap with exon %s" % (
+                    start, end, exon)
+            return 0, exon
+
+        if best_distance > distance:
+            best_distance = distance
+            best_exon = exon
+
+    return best_exon, best_exon
```
timodonnell: best_distance, best_exon ?

varcode/core_logic.py
```diff
@@ -381,6 +383,78 @@ def infer_exonic_effect(variant, transcript):
...
+    best_exon = None
+    for exon in exons:
+        if exon.start > end:
+            # distance from end of variant to start of exon
+            # (when variant starts before the exon)
+            distance = exon.start - end
+        elif exon.end < start:
+            distance = start - exon.end
+        else:
+            assert exon.start <= end and exon.end >= start, \
```
timodonnell: Given the above conditionals, this assert can't fail, but I guess that's what asserts are for

varcode/core_logic.py
```diff
@@ -381,6 +383,78 @@ def infer_exonic_effect(variant, transcript):
...
         transcript)
 
+def find_nearest_exon(start, end, exons):
+    """
+    Finds nearest exon to an intronic variant, returns the
+    distance to that exon, along with the Exon object itself.
+    """
+    best_distance = float("inf")
+    best_exon = None
+    for exon in exons:
```
timodonnell: i'd consider putting this logic in a distance_between_intervals(start1, end1, start2, end2) function

varcode/variant_collection.py
```diff
@@ -125,29 +125,33 @@ def variant_effects(self, raise_on_error=True):
...
             for variant
             in self.records
         ]
 
-    def effects_to_string(self):
+    def effects_to_string(self, *args, **kwargs):
         """
         Create a long string with all transcript effects for each mutation,
         grouped by gene (if a mutation affects multiple genes).
         """
```
timodonnell: a note in the docstring about where the args and kwargs are sent would be helpful

varcode/variant_collection.py
```diff
@@ -163,8 +167,8 @@ def effects_to_string(self):
                 lines.append("  Highest Priority Effect: %s" % best)
         return "\n".join(lines)
 
-    def print_effects(self):
+    def print_effects(self, *args, **kwargs):
         """
         Print all variants and their transcript effects (grouped by gene).
         """
```
timodonnell: similarly re args and kwargs

varcode/core_logic.py
```diff
@@ -381,6 +383,78 @@ def infer_exonic_effect(variant, transcript):
...
+            return 0, exon
+
+        if best_distance > distance:
+            best_distance = distance
+            best_exon = exon
+
+    return best_exon, best_exon
+
+
+def infer_intronic_effect(variant, transcript, nearest_exon, distance_to_exon):
```
timodonnell: nit: I'd consider not passing around the distance_to_exon value. it's cheap to compute given the variant and the exon, and passing it around opens an opportunity for inconsistency (would want to have a helper function to do this though)

timodonnell:
module comments LGTM

varcode/core_logic.py
iskandr: :+1:

varcode/core_logic.py
iskandr: This is more of a documenting assert, to remind the reader what the else branch means.

varcode/core_logic.py
iskandr: Good catch, thanks. I think this calls for a unit test or two.

varcode/core_logic.py
iskandr: Added a helper method to all Locus objects (Gene, Transcript, Exon) to make this easy to compute.

iskandr:
I moved `find_nearest_exon` into PyEnsembl, see: https://github.com/hammerlab/pyensembl/pull/39/

varcode/variant_collection.py

varcode/variant_collection.py

PR #20, opened 2015-02-20
Associate EnsemblRelease with each Variant object

iskandr:
- every Variant must be constructed with an explicit `EnsemblRelease` (this is mildly annoying/inelegant but simplifies a lot of other code)
- moved a lot of the transcript effect inference from  `core_logic` onto `Variant`
- renamed `VariantEffect` to `Annotation`
- added a `Variant.annotate()` method, which replaces the need for `VariantAnnotator`
- got rid of the generic `load_variants` function, now call `load_vcf` or `load_maf` explicitly
- since each Variant carries its own `EnsemblRelease`, it's now possible for variants from multiple references to be in the same `VariantCollection`. I've seen a TCGA-derived MAF which mixes NCBI builds, so this might be desirable (it also gives meaning to taking the union of two variant collections aligned to different references). However, the potential for confusion here is suspicious.


test/test_cosmic_mutations.py
```diff
@@ -9,11 +10,11 @@
     FrameShiftTruncation,
 )
 
-annot = VariantAnnotator(75)
+ensembl = EnsemblRelease(75)
```
timodonnell: I'd suggest renaming this default_ensembl

iskandr: Can you say more about why that's a better name?

varcode/variant.py
```diff
@@ -46,3 +114,240 @@ def short_description(self):
...
+    def coding_genes(self):
+        """
+        Protein coding transcripts
+        """
+        return [
+            gene for gene in self.genes()
+            if is_coding_biotype(gene.biotype)
+        ]
+
+    def annotate(self, raise_on_error=True):
```
timodonnell: to me, annotate() sounds like a mutator. I'd suggest renaming this annotation() and memoizing it.

timodonnell: thinking more, I could go either way on memorizing it actually, whichever you think is best

varcode/variant.py
```diff
@@ -46,3 +114,240 @@ def short_description(self):
...
+    def transcript_effect(self, transcript):
+        """
+        Generate a transcript effect (such as FrameShift) by applying a
+        genomic variant to a particular transcript.
+
+        Parameters
+        ----------
+
+        transcript :  Transcript
+            Transcript we're going to mutate
```
timodonnell: my first reading of the word "mutate" was in the software sense, not the biological sense. Might change this to the more wordy but clearer "Transcript we're going to apply the mutation to". May also want to be clear about this by changing the main method description above to something like "Return the transcript effect (such as FrameShift) from applying a genomic variant to a particular transcript."

varcode/variant.py
```diff
@@ -46,3 +114,240 @@ def short_description(self):
...
+        return effect_class(
+                variant=self,
+                transcript=transcript,
+                nearest_exon=nearest_exon,
+                distance_to_exon=distance_to_exon)
+
+
+    def offset_on_transcript(self, transcript):
+        """
+        Given a Variant (with fields `ref`, `alt`, `pos`, and `end`),
```
timodonnell: suggested minor cleanup:

Given a Variant (with fields `ref`, `alt`, `pos`, and `end`), compute the offset of the variant position relative to the position and direction of the transcript. Returns a tuple (offset, ref, alt), where `ref` and `alt` are this variant's `ref` and `alt` nucleotide strings with shared prefixes and suffixes removed and the bases reversed if the transcript is on the backward strand.

For example,

```
>> variant = Variant(ref="ATGGC", alt="TCGC", pos=3000, end=3005)
>> variant.offset_on_transcript(Transcript(start=2000, end=4000))
(1002, "G", "C")
```

varcode/variant.py
```diff
@@ -46,3 +114,240 @@ def short_description(self):
...
+        # bump the offsets and make the strings disjoint (ref="", alt="C")
+        ref, alt, prefix, _ = trim_shared_flanking_strings(ref, alt)
+        n_same = len(prefix)
+        start_offset_with_utr5 = interval_offset_on_transcript(
+            self.pos, self.end, transcript)
+        offset =  start_offset_with_utr5 + n_same
+        return offset, ref, alt
+
+    def exonic_transcript_effect(self, transcript):
+        start_offset_with_utr5, ref, alt = self.offset_on_transcript(transcript)
```
timodonnell: docstring?

varcode/variant.py
iskandr: What do you think of renaming `Annotation` (which is vague) to
`TranscriptEffectCollection` and changing the accessor method to
`variant.transcript_effects()`?

(we can add memoization if/when we find ourselves calling these methods
repeatedly and caring about the peformance)

varcode/transcript_helpers.py
```diff
@@ -0,0 +1,63 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import print_function, division, absolute_import
+
+def interval_offset_on_transcript(start, end, transcript):
+    """
+    Given a variant and a particular transcript,
```
timodonnell: this does not appear to take a Variant

varcode/variant.py
```diff
@@ -1,39 +1,107 @@
...
+)
 
-from memoized_property import memoized_property
+from pyensembl import Transcript, find_nearest_locus, EnsemblRelease
 from pyensembl.locus import normalize_chromosome
+from pyensembl.biotypes import is_coding_biotype
 
 class Variant(object):
-    def __init__(self, contig, pos, ref, alt, info=None):
+    def __init__(self, contig, pos, ref, alt, ensembl, info=None):
```
timodonnell: ensembl is not actually optional in this parameter list

varcode/variant.py
```diff
@@ -1,39 +1,107 @@
...
+            won't work.
+        """
         self.contig = normalize_chromosome(contig)
-        self.pos = int(pos)
         self.ref = normalize_nucleotide_string(ref)
         self.alt = normalize_nucleotide_string(alt)
+        self.pos = int(pos)
+        self.end = self.pos + len(self.ref) - 1
+
+        if not isinstance(ensembl, EnsemblRelease):
```
timodonnell: not treating ensembl as optional

varcode/variant.py
```diff
@@ -46,3 +114,240 @@ def short_description(self):
...
+                            error)
+                    errors[transcript] = error
+            gene_transcript_effects_groups[gene_id] = effects
+
+        # construct Gene objects for all the genes containing
+        # all the transcripts this variant overlaps
+        overlapping_genes = [
+            self.ensembl.gene_by_id(gene_id)
+            for gene_id
+            in overlapping_transcript_groups.keys()
```
timodonnell: nit: no need for .keys()

varcode/variant.py
```diff
@@ -46,3 +114,240 @@ def short_description(self):
...
+        if distance_to_exon > 0:
+            return self.intronic_transcript_effect(
+                transcript=transcript,
+                nearest_exon=nearest_exon)
+
+        # TODO: exonic splice site mutations
+        return self.exonic_transcript_effect(transcript)
+
+
+    def intronic_transcript_effect(self, transcript, nearest_exon):
```
timodonnell: consider making this "private" (i.e. prefix an underscore to the name)

varcode/variant_collection.py
```diff
@@ -110,49 +67,55 @@ def __str__(self):
...
 
     def drop_duplicates(self):
         """
         Create a new VariantCollection without any duplicate variants.
         """
-        return self.clone(set(self.variants))
+        return self.clone_metadata(set(self.variants))
 
-    def variant_effects(self, raise_on_error=True):
+    def annotations(self, raise_on_error=True):
```
timodonnell: this method may be unnecessary? I'm fine to call .annotation myself on each variant in a list

varcode/variant_collection.py
```diff
@@ -110,49 +67,55 @@ def __str__(self):
...
-                variant=variant,
+           variant.annotate(
                 raise_on_error=raise_on_error)
             for variant
             in self.variants
         ]
 
+    def high_impact_variants(self, *args, **kwargs):
+        """
+        Returns a list of VariantEffect objects for variants predicted
```
timodonnell: didn't VariantEffect get renamed Annotation?

varcode/variant_collection.py
```diff
@@ -110,49 +67,55 @@ def __str__(self):
...
+        to have a significant impact on some transcript's function.
+
+        All arguments are passed on to variant_effects(*args, **kwargs).
+        """
+        effects = self.variant_effects(*args, **kwargs)
```
timodonnell: missing return statement? is this method necessary?

varcode/variant_collection.py
```diff
@@ -110,49 +67,55 @@ def __str__(self):
     def __repr__(self):
         return str(self)
 
-    def clone(self, new_variants=None):
+    def clone_metadata(self, new_variants):
```
timodonnell: i'd suggest having one method called `clone` defined immediately before or after __init__ (so it's easy to keep the arguments in agreement) that takes exactly the same arguments as __init__ but with all arguments optional. Any arguments specified to that function will be used in the returned VariantCollection, and any arguments omitted will default to the values in the current instance. Otherwise having multiple clone methods and keeping track of what's considered "metadata" as new fields are added in the future seems confusing to me

varcode/variant_collection.py
iskandr: Sorry, this method wasn't done, I'll finish it up.

varcode/variant_collection.py
iskandr: Funny, I just changed it from `clone` to `clone_metadata` since Tavi felt like it's strange to allow the possibility that `x != x.clone(some_args)`.

varcode/variant.py
iskandr: :+1:

varcode/variant.py
iskandr: :+1:

varcode/variant.py
iskandr: The docstring parameter is wrong, it's not optional.

varcode/variant_collection.py
```diff
@@ -110,49 +67,55 @@ def __str__(self):
...
+
+    def clone(self):
+        return self.clone_metadata(list(self.variants))
 
     def drop_duplicates(self):
         """
         Create a new VariantCollection without any duplicate variants.
         """
-        return self.clone(set(self.variants))
+        return self.clone_metadata(set(self.variants))
```
timodonnell: I think there's a bug floating around here, if we're now returning an instance that has a `set` of variants instead of a `list`, since the `__eq__` method checks equality by iterating through the zip'd variants, which assumes a particular sort order. I'm confused at why the `__eq__` method does that actually. My suggestion is to (1) drop the `__eq__` method -- I think users will just want to compare the .variants using usual python semantics, instead of depending on the intrinsically shaky semantics of equality in VariantCollection (e.g. what if the filename is different, is that a different collection? I just looked at that code an have already forgotten) and (2) either remove this `drop_duplicates` method, or modify it to create a VariantCollection with a `list` of variants instead of a set (this will require removing duplicates in a way that preserves sort order)

varcode/variant_collection.py
```diff
@@ -161,10 +124,10 @@ def effects_to_string(self, *args, **kwargs):
         Arguments are passed on to variant_effects(*args, **kwargs).
         """
         lines = []
-        for variant_effect in self.variant_effects(*args, **kwargs):
+        for annotation in self.annotations(*args, **kwargs):
```
timodonnell: header comment needs to be updated for new "annotation" language. Consider renaming method to something like `annotations_summary`?

timodonnell: Also I'd advocate for removing the `print_effects` method. I can print this string if I need to.

varcode/vcf.py
```diff
@@ -34,19 +43,39 @@ def load_vcf(filename, reference_path_field='reference'):
...
         Variant(
-            x.CHROM, x.POS,
-            x.REF, x.ALT[0].sequence,
-            x.INFO)
+            contig=x.CHROM,
+            pos=x.POS,
+            ref=x.REF,
+            alt=x.ALT[0].sequence,
+            info=x.INFO,
+            ensembl=ensembl)
```
timodonnell: I'm pretty sure what you have is correct (below), but just double checking -- we don't have to compare x.FILTER to "." also, do we?

test/test_cosmic_mutations.py
timodonnell: It's used only as a default parameter, right? Otherwise it looks to me like we're declaring that this module uses a particular ensembl release

timodonnell: ah wait sorry, nvm. somehow didn't realize this was in a test method.

test/test_vcf.py
```diff
@@ -12,33 +12,30 @@
...
-        "Expected hg19 reference, got %s" % (variants.reference_path,)
+    variants = load_vcf(VCF_FILENAME)
     # after normalization, hg19 should be remapped to GRCh37
-    assert variants.reference_name == "GRCh37"
+    assert variants.reference_names() ==  { "GRCh37" }
 
 def test_vcf_number_entries():
     # there are 14 mutations listed in the VCF, make sure they are all parsed
-    variants = load_variants(VCF_FILENAME)
+    variants = load_vcf(VCF_FILENAME)
```
timodonnell: regarding the assert one line below: not a big deal, and no need to change this in this PR, but I'd suggest using something like the numpy testing library (numpy.testing) to avoid having to write out assertion failure strings like this

varcode/variant.py
timodonnell: Sure, I'd be fine with this change. Consider also: `TranscriptModification` instead of `TranscriptEffect`. I don't love either one very much, but the "modification" language at least resolves the effect-on-transcript vs. effect-of-transcript ambiguity

varcode/variant_collection.py
timodonnell: Gotcha. I'd be fine with any name for this method, as long as there is only one method

timodonnell:
Well hope you don't mind the pile of comments, but besides all that, LGTM

varcode/variant_collection.py
iskandr: With the new names, what do you think of `effect_summary` for the to_string method?

timodonnell: Works for me!

varcode/vcf.py
iskandr: PyVCF takes care of it:

```
   if filt == '.':
            filt = None
```

test/test_cosmic_mutations.py
iskandr: :+1:

test/test_vcf.py
iskandr: You can use `nose.tools.eq_` but sometimes I find it more informative to have a string explaining where the value came from.

varcode/variant.py
iskandr: :+1:

varcode/variant.py
iskandr: Thanks for the cleanup, the logic is messy but I wasn't sure how else to organize/describe it.

varcode/variant.py
iskandr: :+1:

varcode/variant_collection.py

varcode/variant_collection.py
iskandr: In its newer incarnation, this method gives options that are annoying to recreate manually:

```
    def variant_effects(
            self,
            min_effect_class=None,
            only_coding_transcripts=False,
            raise_on_error=True):
```

PR #24, opened 2015-02-24
Variant collection tweaks

timodonnell:
* Variants are now stored in a set, not a list. The VariantCollection constructor can take any iterable of variants and will convert it to a set.
 * The `__str__` method now gives a short summary with just the variant count. The original `__str__` method, which gives a line for each variant, has been moved into the `variant_summary` method.
 * The `reference_names` and `gene_counts` methods have been converted to memoized properties.

 * Remove VariantCollection.drop_duplicates since variants are now stored in a set.
 * Make Variant instances comparable. They are ordered by locus.
 * Update tests


test/test_drop_duplicates.py
```diff
@@ -1,27 +0,0 @@
...
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
-from pyensembl import EnsemblRelease
-from varcode import Variant, VariantCollection
-
-def test_drop_duplicates():
```
iskandr: I'm not sure about dropping this test entirely, we still probably want to ensure that the expect behavior happens when you initialize a variant collection with duplicates (though now that expected behavior is just the second assert below).

test/test_maf.py
```diff
@@ -7,12 +7,12 @@ def test_maf():
...
     expected_variants = [
         Variant(1, 1650797, "A", "G", ensembl),
-        Variant(1, 231401797, "A", "C", ensembl),
         Variant(1, 23836447, "C", "A", ensembl),
+        Variant(1, 231401797, "A", "C", ensembl),
         Variant(11,124617502, "C", "G", ensembl),
     ]
     eq_(len(variant_collection_from_maf), len(expected_variants))
-    for v_expect, v_maf in zip(expected_variants, variant_collection_from_maf):
+    for v_expect, v_maf in zip(expected_variants, sorted(variant_collection_from_maf)):
```
iskandr: Does using `set` inside `VariantCollection` screw up the given order? If so, maybe the `sorted` should go on the `__iter__` method of `VariantCollection`.

iskandr:
Other than (1) adding a test which explicitly expects dropping duplicates by default and (2) moving `sorted` into `VariantCollection.__iter__`, LGTM

test/test_maf.py
timodonnell: Yeah, using a `set` results in an undefined order for iteration.

Agree with your suggestion. Sorting is fast enough that I don't see it mattering to do it every time we iterate through the collection, even for large collections. If we do hit performance issues, may want to move to a sorted set implementation.

test/test_drop_duplicates.py
timodonnell: Good call

timodonnell:
Updated to address comments, but also added a new commit (d1cd61693484bf614231038b63ce9f93b3c10c2f) making a small change to Variant constructor. Please have a look at that @iskandr

PR #27, opened 2015-02-25
Remove pyfaidx

iskandr:
I wasn't using it in any interesting way, can be replaced with `Bio.Seq.reverse_complement` instead.


timodonnell:
I was hesitant to file that annoying issue, but glad I did -- this is a good improvement! LGTM

Issue #22, opened 2015-02-23
handle multiallelic variants


timodonnell:
Currently, `vcf.load_vcf` just takes the first alternative allele.

iskandr:
What do you think is the correct way to handle a multi-allelic variant? How do they differ from having two variants at the same locus?

timodonnell:
Yeah, I think having the `load_vcf` function return separate variants for each allele in multi-allelic site would work.

PR #29, opened 2015-03-03
parse multiple alleles into distinct Variant records

iskandr:
Fixes https://github.com/hammerlab/varcode/issues/22.

Also fixes bug from the last PEP8 PR which removed `reverse_complement` from `string_helpers`. 

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/29)

### DIFF
--- a/varcode/vcf.py
+++ b/varcode/vcf.py
@@ -65,19 +65,21 @@ def load_vcf(
 
     ensembl = EnsemblRelease(release=ensembl_release)
 
-    # We ignore "no-call" variants, i.e. those where X.ALT = [None].
-    variants = [
-        Variant(
-            contig=x.CHROM,
-            pos=x.POS,
-            ref=x.REF,
-            alt=x.ALT[0].sequence,
-            info=x.INFO,
-            ensembl=ensembl)
-        for x in vcf_reader
-        if x.ALT[0] and (not only_passing or not x.FILTER or x.FILTER == "PASS")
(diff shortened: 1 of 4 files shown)


varcode/vcf.py
```diff
@@ -65,19 +65,21 @@ def load_vcf(
...
-            ref=x.REF,
-            alt=x.ALT[0].sequence,
-            info=x.INFO,
-            ensembl=ensembl)
-        for x in vcf_reader
-        if x.ALT[0] and (not only_passing or not x.FILTER or x.FILTER == "PASS")
-    ]
-
+    variants = []
+    for record in vcf_reader:
```
timodonnell: I suppose this has always been the case for varcode, but this just made me think about it: we're throwing away the genotype information. Even for a regular non-multiallelic variant, I don't think I have a way to see if it was called a het or hom. Would it make sense to retain that in the variant? Or we could potentially address that in a later PR.

timodonnell:
LGTM

varcode/vcf.py
iskandr: Yep, currently we're only attaching the INFO dict to variants and not the sample info. What do you think is the right way to handle e.g. genotypes? Should we make a nested sample_name -> key -> value dictionary of dictionaries? That seems messy but I'm not sure what else to do. In any case, that should definitely be a separate PR.

PR #34, opened 2015-03-08
Fix maf parsing

iskandr:
This pull request started with just fixing the parsing logic in `maf.py` to deal with some of the edge cases of TCGA's various MAF files (spaces in values, non-standard capitalization of columns). In the process of adding unit tests I realized I wasn't handling `StopLoss` variants at all in `coding_effect.py`, which may have been responsible for some errors I've seen while parsing larger germline VCF files. 

The branch eventually came to encompass a refactoring of `Variant`, addition of `Intragenic` & `Intragenic` effects which don't require transcripts, and removal of `VariantEffectCollection` (in favor of a simple list of variants and more effect-related helpers on the `Variant` itself).

One of the MAF effect tests still doesn't work, but I guess there's a lot to start reviewing that probably won't be affected by remaining changes. 

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/34)


varcode/variant.py
```diff
@@ -175,71 +197,86 @@ def coding_genes(self):
...
+                        else:
+                            logging.warn(
+                                "Encountered error annotating %s for %s: %s",
+                                self,
+                                transcript,
+                                error)
+        return effects
+
+    @memoize
+    def transcript_effect_dict(self, *args, **kwargs):
```
timodonnell: At first glance it's not clear why we want this method in addition to effects() that returns a VariantEffectCollection. Maybe consider getting rid of VariantEffectCollection, moving its methods into Variant, and having `variant.effects()` return the dict that is currently returned by `transcript_effect_dict`?

varcode/variant.py
```diff
@@ -175,71 +197,86 @@ def coding_genes(self):
...
         """
-        Return the transcript effect (such as FrameShift) that results from
+        return {
+            effect.transcript.id: effect
+            for effect in self.effects(*args, **kwargs)
+            if isinstance(effect, TranscriptMutationEffect)
+        }
+
+    @memoize
+    def summary_effect(self, *args, **kwargs):
```
timodonnell: The name `summary_effect` to me makes it sound like it's somehow summarizing the effects across multiple transcripts, maybe consider changing to `top_effect` or similar

varcode/variant_collection.py
```diff
@@ -92,58 +94,55 @@ def _clone_metadata(self, new_variants):
             variants=new_variants,
             original_filename=self.original_filename)
 
+    @memoize
+    def high_impact_effects(self, *args, **kwargs):
```
timodonnell: I think the `*args` and `**kwargs` forwarding is confusing, and I'd expect this method to take a `min_priority` argument that defaults to `Substitution`. Some ideas:

 * Take an explicit `min_priority` and `raise_on_error` arguments.
 * Have a `VariantCollection.raise_on_annotation_error` boolean that user can set to indicate the whether to raise annotation errors on any method call for that object.
 * Have a global `varcode.RAISE_ON_ANNOTATION_ERROR` boolean that anything can look at. If we care about thread saftey (not saying we should), could have a `varcode.set_raise_on_annotation_error(value)` and `varcode.get_raise_on_annotation_error()` function that saves this value in thread local storage.

varcode/variant.py
iskandr: I was trying to think of a better name, `top_effect` works for me.

varcode/variant_collection.py
iskandr: I'm going to leave argument forwarding for now since the set of globally relevant parameters hasn't stabilized yet. I do agree that eventually it would be nice to have explicit parameters on each method. 

Changing the `raise_on_error` parameter into a `VariantCollection.raise_on_annotation_error` property sounds good to me, adding it as an issue: https://github.com/hammerlab/varcode/issues/36

varcode/variant.py
iskandr: `VariantEffectCollection` is gone on this branch, and all of its methods and properties (like this one) have migrated onto `VariantCollection`.

PR #35, opened 2015-03-10
Typechecks and test fixes

timodonnell:
* Replace reference to pyensembl type_checks module with typechecks package
 * Fix nondeterministic test failure in test_vcf.py
 * Add test utility module (`__init__.py` in the test directory) with `data_path` function for getting the absolute path to test resources
 * Modify tests to use `data_path`, instead of assuming that the current working directory is the 'test' directory. This makes it possible to run `nosetests` either in the root `varcode` directory or in the `test` directory.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/35)

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -1,0 +1,12 @@
+'''
+Utility functions for tests.
+'''
+
+import os
+
+def data_path(name):
+    '''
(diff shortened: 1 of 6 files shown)


iskandr:
LGTM

PR #37, opened 2015-03-18
Test problematic variants

iskandr:
* added unit test module for "problematic" variants (anything which raises an exception when annotating variants in the wild, e.g. from TCGA MAF files)
* `Variant` now trims any shared suffix or prefix strings from `ref` and `alt` and adjusts the variant position accordingly. The position field has been renamed to `start`. The original variant fields are preserved as `Variant.original_ref`, `Variant.original_alt`, `Variant.original_start`. For example, if a VCF contains "chr1 10 CAG CATG", then the corresponding `Variant` object will have `ref = ""`, `alt = "T"`, `start = 11`, `original_ref = "CAG"`, `original_alt = "CATG"`, `original_start = 10`. 
* Renamed `top_priority_transcript_effect` to `top_priority_effect` (since we now have effects like `Intergenic` which aren't tied to a particular transcript)
* Added `ExonLoss` as an effect
* Reorganized the effect annotation logic in `Variant` to deal with variants which span feature boundaries (e.g. deletions which affect both an exon and its neighboring intron). 
* renamed `VariantCollection.variant_effects` to just `VariantCollection.effects`
* renamed `VariantCollection.variant_summary` to `VariantCollection.summary_string`
* added unit tests for misc. `VariantCollection` methods

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/37)

### DIFF
--- a/test/test_problematic_variants.py
+++ b/test/test_problematic_variants.py
@@ -1,0 +1,48 @@
+"""
+Any variants which are encountered in the wild and either cause Varcode
+to crash or return an incorrect annotation should be added to this
+test module.
+"""
+
+from pyensembl import EnsemblRelease
+from varcode import Variant
+
+ensembl75 = EnsemblRelease(75)
+
+# variants which have previously resulted in raised exceptions
+# during effect annotation
+should_not_crash_variants = [
+    # error message:
+    # "Couldn't find position 92979124 on any exon of ENST00000540033"
+    Variant(
+        contig=1,
+        pos=92979092,
+        ref="ATATATATATATATATATATATATATATATATG",
+        alt="A",
+        ensembl=ensembl75),
+    # error message:
+    # "Expect non-silent stop-loss variant to cause longer variant protein"
+    # "" but got len(original) = 653, len(variant) = 653"
+    Variant(
+        contig=1,
+        pos=167385324,
+        ref="TAA",
+        alt="T",
+        ensembl=ensembl75),
+    # error message:
+    # "Variant which span 5' UTR and CDS not supported"
+    Variant(
+        contig=19,
+        pos=44351166,
+        ref="GGGAGAT",
+        alt="G",
+        ensembl=ensembl75)
+]
+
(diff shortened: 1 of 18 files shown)


test/test_problematic_variants.py
```diff
@@ -0,0 +1,88 @@
...
+        ensembl=ensembl75),
+]
+
+def try_effect_annotation(variant):
+    effect = variant.top_effect()
+    assert effect is not None
+
+def test_crashing_variants():
+    for variant in should_not_crash_variants:
+        yield (try_effect_annotation, variant)
```
timodonnell: will this code actually run? will nose call next() automatically on tests that return generators?

timodonnell: nvm, I googled, and the answer is yes! good to know http://nose.readthedocs.org/en/latest/writing_tests.html

varcode/coding_effect.py
```diff
@@ -81,33 +53,53 @@ def infer_coding_effect(
...
             variant,
             ref,
-            cds_offset,
+            transcript_offset,
             transcript,
-            cds_ref)
+            transcript_ref)
+
+    cds_start_offset = min(transcript.start_codon_spliced_offsets)
+    cds_stop_offset = max(transcript.stop_codon_spliced_offsets)
```
timodonnell: you probably have it right, but just want to double check: are these min and max what you want even if it's on the reverse strand?

timodonnell:
I didn't check all the logic here (really unit tests are the only way), and sounds like much of this will change if you move to the `ref_position_before` and `ref_position_after` way of localizing variants, but from looking at what you have so far I don't see any major flaws

varcode/coding_effect.py
iskandr: All the offsets on `Transcript` (like `start_codon-spliced_offsets`) are relative to the direction off the transcript regardless of which strand it's on.

iskandr:
Did anything stand out as particularly in need of more/any unit tests?

PR #39, opened 2015-03-24
Small coding effect refactoring and fixes

iskandr:
* added `AlternateStartCodon` effect for changes like c.1ATG>TTG
* factored out `translate` function in `coding_effect`
* changed logic of translation to treat alternate start codons as Methionines

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/39)

### DIFF
--- a/varcode/coding_effect.py
+++ b/varcode/coding_effect.py
@@ -45,6 +45,9 @@ def translate(cds_seq):
     first residues since the CDS we're translating might have been affected
     by a start loss mutation.
 
+    The sequence may include the 3' UTR but will stop translation at the first
+    encountered stop codon.
+
     Parameters
     ----------
     cds_seq : str or BioPython Seq
--- a/varcode/coding_effect.py
+++ b/varcode/coding_effect.py
@@ -53,9 +53,6 @@ def translate(cds_seq):
     cds_seq : str or BioPython Seq
         cDNA coding sequence
 
-    transcript : PyEnsembl.transcript
-        Transcript from which we got cds_seq, used for error messages.
-
     Returns BioPython Seq of amino acids
     """
     cds_seq = Seq(cds_seq)
--- a/varcode/coding_effect.py
+++ b/varcode/coding_effect.py
@@ -191,8 +191,13 @@ def infer_coding_effect(
         elif aa_pos == len(original_protein):
             aa_ref = "*"
         elif aa_pos > len(original_protein):
-            logging.warn("How did we get aa_pos = %d when len(protein) = %d?",
-                aa_pos, len(original_protein))
+            logging.warn(
+                ("Unexpected aa_pos = %d  for len(protein) = %d"
+                 " in 3' UTR of %s for %s"),
+                aa_pos,
+                len(original_protein),
+                transcript,
+                variant)
             aa_ref = "?"
         return Silent(
             variant,
--- a/varcode/coding_effect.py
+++ b/varcode/coding_effect.py
@@ -22,17 +22,54 @@ from .effects import (
     Substitution,
     ComplexSubstitution,
     PrematureStop,
+    AlternateStartCodon,
     StartLoss,
     StopLoss,
     FrameShift,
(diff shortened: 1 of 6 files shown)


varcode/coding_effect.py
```diff
@@ -22,17 +22,54 @@
...
+STOP_CODONS = set(CodonTable.standard_dna_table.stop_codons)
+
+def translate(cds_seq):
+    """
+    Parameters
+    ----------
+    cds_seq : str or BioPython Seq
+        cDNA coding sequence
+
+    transcript : PyEnsembl.transcript
```
timodonnell: this does not appear to take this argument

varcode/coding_effect.py
```diff
@@ -22,17 +22,54 @@
...
-from Bio.Seq import Seq, CodonTable
+from Bio.Seq import Seq
+from Bio.Data import CodonTable
+
+START_CODONS = set(CodonTable.standard_dna_table.start_codons)
+STOP_CODONS = set(CodonTable.standard_dna_table.stop_codons)
+
+def translate(cds_seq):
+    """
+    Parameters
```
timodonnell: a one sentence description of the logic this function follows might be helpful. This is probably second nature to you, but from reading this description it's not immediately clear to me what exactly `cds_seq` is -- e.g. does it include the UTR's?

varcode/coding_effect.py
```diff
@@ -22,17 +22,54 @@
...
+    # passing cds=False since we may want to deal with premature
+    # stop codons
+    protein = truncated_cds_seq.translate(to_stop=True, cds=False)
+
+    if protein[0] != "M" and cds_seq[:3] in START_CODONS:
+        # TODO: figure out when these should be made into methionines
+        # and when left as whatever amino acid they normally code for
+        # e.g. Leucine start codons
+        # See: DOI: 10.1371/journal.pbio.0020397
+        return "M" + protein[1:]
```
timodonnell: From a quick reading of things like 10.1126/science.1224439 and 10.1126/science.1220270 , it seems like CUG start codons will truly translate as L, whereas other cryptic start codons may initiate translation by mis-pairing with methionine tRNA's, so should be translated as M. Does that fit with what you've seen?

If that's true, it seems like this is the wrong behavior for CUG start codons (i.e. they should be left as L). Seems reasonable to also leave this as a TODO, though.

varcode/coding_effect.py
```diff
@@ -162,37 +181,50 @@ def infer_coding_effect(
...
+                original_start_codon,
+                variant_start_codon)
+        elif aa_pos < len(original_protein):
             aa_ref = original_protein[aa_pos]
         elif aa_pos == len(original_protein):
             aa_ref = "*"
         elif aa_pos > len(original_protein):
-            logging.warn("How did we get aa_pos = %d when len(protein) = %d?",
-                aa_pos, len(original_protein))
+            if cds_seq[:len(original_protein) * 3 + 3].endswith("TGA"):
```
timodonnell: Maybe add a quick comment giving the rationale (and maybe a reference) for this selenocysteine logic?

timodonnell:
Modulo comments LGTM

varcode/coding_effect.py
iskandr: It's unclear to me what to do with CUG since some papers also show it coding for methionine. 

See: http://www.jbc.org/content/285/7/4595.full

iskandr: Also:

>...ACG, CUG, AUU, AUC, and AUA can all initiate protein synthesis to various degrees in vivo and in vitro (39). These mutant initiation codons differ from AUG by a single nucleotide. Because they use the wild-type initiator tRNA for initiation, protein synthesis in all of these cases is initiated with methionine

(http://mcb.asm.org/content/18/9/5140.full)

I think finding that CUG can initiate in vivo with a Leucine is novel but I don't know where it fits in relative to CUG methionine starts (which is more common or likely?)

PR #45, opened 2015-03-26
Use transcript protein sequence

iskandr:
Overall motivation: move toward translating as few codons as necessary for each annotation (instead of always translating the full original & mutated proteins)

* Use `transcript.protein_sequence` instead of translating original protein
* Created special case for SNVs in `coding_effect` (plan to generalize this as a cleaner logic for all in-frame coding effects), translates only single affected codon
* Uses PyEnsembl 0.6.2

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/45)

### DIFF
--- a/varcode/coding_effect.py
+++ b/varcode/coding_effect.py
@@ -15,6 +15,9 @@
 from __future__ import print_function, division, absolute_import
 import logging
 
+# from Bio.Seq import Seq
+from Bio.Data import CodonTable
+
 from .effects import (
     Silent,
     Insertion,
@@ -32,11 +35,27 @@ from .effects import (
 from .mutate import substitute, insert_after
 from .string_helpers import trim_shared_flanking_strings
 
-# from Bio.Seq import Seq
-from Bio.Data import CodonTable
-
+DNA_CODON_TABLE = CodonTable.standard_dna_table.forward_table
 START_CODONS = set(CodonTable.standard_dna_table.start_codons)
 STOP_CODONS = set(CodonTable.standard_dna_table.stop_codons)
+
+def translate_codon(codon, aa_pos):
+    """Translate a single codon into a single amino acid or stop '*'
+
+    Parameters
+    ----------
+    codon : str
+        Expected to be of length 3
+    aa_pos : int
+        Codon/amino acid offset into the protein (starting from 0)
+    """
+    # not handling rare Leucine or Valine starts!
+    if aa_pos == 0 and codon in START_CODONS:
+        return "M"
+    elif codon in STOP_CODONS:
+        return "*"
+    else:
+        return DNA_CODON_TABLE[codon]
 
 def translate(cds_seq):
     """Translates cDNA coding sequence into amino acid protein sequence.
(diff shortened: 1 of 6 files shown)


varcode/coding_effect.py
```diff
@@ -32,12 +35,28 @@
...
 START_CODONS = set(CodonTable.standard_dna_table.start_codons)
 STOP_CODONS = set(CodonTable.standard_dna_table.stop_codons)
 
+def translate_codon(codon, aa_pos):
+    """Translate a single codon into a single amino acid or stop '*'
+
+    Parameters
+    ----------
+    codon : str
+        Expected to be of length 3
```
timodonnell: might as well `assert` this?

varcode/coding_effect.py
```diff
@@ -32,12 +35,28 @@
...
+    """
+    # not handling rare Leucine or Valine starts!
+    if aa_pos == 0 and codon in START_CODONS:
+        return "M"
+    elif codon in STOP_CODONS:
+        return "*"
+    else:
+        return DNA_CODON_TABLE[codon]
+
 def translate(cds_seq):
```
timodonnell: super nitpicky nit: maybe rename `cds_seq` to something a little clearer like `coding_sequence` or even just `cds`.  If `cds` stand for "coding sequence", then `cds_seq` would be a sequence of coding sequences?

Also perhaps clarify the significance of it being a `cDNA coding sequence` in the comment? Does that mean that it's the reverse complement of the mRNA transcript, or is that meant to imply some more subtle restriction on what I can pass to this function?

timodonnell: Realizing that you use `cds_seq` in a number of places. If you prefer that name, that's fine with me.

varcode/util.py
```diff
@@ -60,6 +64,15 @@ def random_variants(count, ensembl_release=MAX_ENSEMBL_RELEASE):
         if transcript.on_backward_strand:
             ref = reverse_complement(ref)
         alt_nucleotides = [x for x in VALID_NUCLEOTIDES if x != ref]
+        if insertions:
+            nucleotide_pairs = [
```
timodonnell: worth testing insertions > length 2?

varcode/util.py
```diff
@@ -60,6 +64,15 @@ def random_variants(count, ensembl_release=MAX_ENSEMBL_RELEASE):
...
+                x + y
+                for x in alt_nucleotides
+                for y in VALID_NUCLEOTIDES
+            ]
+            alt_nucleotides.extend(nucleotide_pairs)
+        if deletions:
+            alt_nucleotides.append("")
```
timodonnell: likewise, maybe include longer deletions?

varcode/util.py
```diff
@@ -28,7 +28,11 @@
 _transcript_ids_cache = {}
 
 
-def random_variants(count, ensembl_release=MAX_ENSEMBL_RELEASE):
+def random_variants(
```
timodonnell: consider setting a random seed (or taking one as an optional argument) so that the tests are deterministic? or does `random` module default to a fixed seed?

timodonnell:
LGTM. I like the random variant annotation testing!

varcode/coding_effect.py
iskandr: Asserting in a second branch that does refactoring on top of this.

varcode/coding_effect.py
iskandr: What do you think of just using `cds`? It's pretty pervasive genomics and I agree that `cds_seq` is redundant.

varcode/util.py
iskandr: Added an issue: https://github.com/hammerlab/varcode/issues/47

varcode/util.py
iskandr: https://github.com/hammerlab/varcode/issues/47

varcode/util.py
iskandr: https://github.com/hammerlab/varcode/issues/48

Also, since I'm currently using the non-determinism of this test to find edge cases, also going to add a commandline script for that purpose: 

https://github.com/hammerlab/varcode/issues/49

varcode/coding_effect.py
timodonnell: Yeah, I have a mild preference for `coding_sequence` but I'd be fine with
`cds`.

PR #61, opened 2015-04-21
include substitution in high priority effects

arahuja:
"at least as high as a missense mutation" would imply to me that includes substitutions, if not I can update the doc instead.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/61)


iskandr:
I just changed this on another branch but no harm merging early.

Issue #63, opened 2015-04-22
AssertionError: aa_ref and aa_alt can't both be empty string


arahuja:
```
v = Variant(contig=1, start=56962223, ref='C', alt='T', ensembl=ensembl)
v.effects()
```
throws that error.

1. I don't know if that AssertionError is that useful from a user point of view, it's complaining about function parameters fairly deep in the code.
2. I believe the error occurs much earlier, but I can't figure out a lot of the logic in `in_frame_coding_effect`

`in_frame_coding_effect` is called with

```
ref= G
alt= A
cds_offset= 935
[...]

Issue #58, opened 2015-04-14
VariantCollection.high_priority_effect != Variant.top_effect


tavinathanson:
I'd expect `Variant` to also have a `high_priority_effect`: the same as `top_effect`, with a min filter.

iskandr:
@arahuja @timodonnell @tavinathanson 

I'm currently reorganizing `Varcode` so `VariantCollection.effects()` returns an `EffectCollection` which has a small `top_effect_per_variant()` that's defined as
```
return {
   variant: top_priority_effect(variant_effects)
   for (variant, variant_effects) in
   self.groupby_variant().items()
```

If you only want to see effects that affect splicing or the protein sequence, you can make a smaller effect collection by calling:

`ec.filter_silent().top_effect_per_variant()`. 

Does this seem reasonable to you guys?

tavinathanson:
What is each member of an `EffectCollection`? A `MutationEffect`?

Does Variant.effects() also return an `EffectCollection`?

iskandr:
Yes and yes.

iskandr:
Yes and yes.

tavinathanson:
Makes sense to me! I don't remember what the previous concerns were (and I think there were concerns?) about having a separate `EffectCollection` class.

iskandr:
@tavinathanson I think it wasn't clear that we needed it, so it was adding boilerplate around things we could do more simply with simple lists and dicts. I think now that the API is emerging from the murkiness of imagination, it actually seems useful to between `VariantCollection` and `EffectCollection`.

iskandr:
Fixed by https://github.com/hammerlab/varcode/pull/66

PR #66, opened 2015-04-24
Created EpitopeCollection, refactored effects, fix assertion failure while annotating silent stop codon

### LINKED ISSUE #64: modifies_coding_sequence is always false
Is 
```
    @property
    def modifies_coding_sequence(self):
        """It's convenient to have a property which tells us:
            1) is this a variant effect overlapping a transcript?
            2) does that transcript have a coding sequence?
            3) does the variant affect the coding sequence?
        """
        return False
```

ever overridden?  I can't find it, so it seems like it is always false?

Example - substitution should return `True`, correct?
```
print effect
print effect.modifies_coding_sequence
```

output
```
Substitution(variant=chr1 g.69532C>T, transcript=OR4F5-001, effect_description=p.H148Y)
False
```

iskandr:
Created an EffectCollection, which is used for filtering and grouping sets of effects.
VariantCollection.effects() now returns an EffectCollection.
Merged functionality of top_effect() and highest_priority_effects into EffectCollection.top_priority_effect()
Created a base class Collection from which VariantCollection and EffectCollection get their simpler methods like `__iter__` and `__len__`
Simplified the effect hierarchy, expanded the ambiguous aa_pos (which means slightly different things for different effect types), into the more explicit aa_mutation_start_offset and aa_mutation_end_offset.

Fixes #64 and #58,  https://github.com/hammerlab/varcode/issues/63

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/66)


varcode/variant_collection.py
```diff
@@ -134,57 +93,23 @@ def effects(
...
-        collection.
+        only_transcript_ids : set, optional
+            If given, then only return effects for transcript IDs present
+            in this set.
         """
-        return {
-            variant.reference_name
-            for variant in self.variants
-        }
+        effect_list = []
```
arahuja: Just a list comprehension?

arahuja:
This looks good to me overall, though prob worth someone else scanning through as well @tavinathanson @timodonnell 

Two general things seem odd to me:
1. effect `__lt__`, sorting the effect by the variant might make sense in some cases, but I'm not sure that is always the expected ordering, especially when effects are usually sorted by some priority?
2. The `start_offset` on `Silent` is a bit weird, since it creates a nonsense interval, and the `start_offset` is inclusive in every case but that.

iskandr:
@arahuja Do you think the effects should be sorted by priority instead?

PR #68, opened 2015-04-28
fix effect collection summary string

arahuja:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/68)


arahuja:
Nevermind this was not against the latest master

PR #72, opened 2015-04-28
Read evidence tweaks

timodonnell:
* fix bug that caused mishandling of soft clipped reads
 * add performance test script
 * some better assertions
 * other minor fixes

Due to the issue https://github.com/pysam-developers/pysam/issues/106 these changes require a snapshot pysam until a release is made. Once pysam makes a new release I'll update the dependency to use that.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/72)

### DIFF
--- a/test/read_evidence_performance.py
+++ b/test/read_evidence_performance.py
@@ -16,6 +16,11 @@ import time
 import varcode
 import varcode.read_evidence
 from varcode.locus import Locus
+    
+PARSER = argparse.ArgumentParser(usage=__doc__)
+PARSER.add_argument("bam_path")
+PARSER.add_argument("regions", nargs="+")
+PARSER.add_argument("--num-loci-per-region", type=int, default=100)
 
 def parse_region(s):
     (contig, rest) = s.split(":")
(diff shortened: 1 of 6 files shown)


timodonnell:
Tests successful on both python 2.7 and 3.4 for this.

test/read_evidence_performance.py
```diff
@@ -0,0 +1,53 @@
...
+from varcode.locus import Locus
+
+def parse_region(s):
+    (contig, rest) = s.split(":")
+    (start, end) = rest.split("-")
+    return varcode.Locus.from_inclusive_coordinates(
+        contig, int(start), int(end))
+
+def go(argv):
+    parser = argparse.ArgumentParser(usage=__doc__)
```
iskandr: I don't know if this is an insignificant style difference, but I usually define argument parsers at the global scope. Also, why do you need to get `sys.argv[1:]`, wouldn't just `parse_args()` work?

timodonnell: Sure, I can put the parser at global scope.

By doing `sys.argv[1:]` myself I have a function that I could theoretically call from other code, like tests, without assuming anything about the global argv

PR #75, opened 2015-04-29
Fix variant.effects() to always return an EffectCollection

timodonnell:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/75)


iskandr:
LGTM

PR #78, opened 2015-04-29
Variant serialization

### LINKED ISSUE #77: Make Variants pickle-able
Need an implementation of `__getstate__` and `__setstate__` to reduce `EnsemblRelease` objects to their version and then reconstruct that `EnsemblRelease` using `pyensembl.cached_release`. 

See: https://docs.python.org/2/library/pickle.html#pickle-protocol

timodonnell:
Adds support for serializing Variant instances as pickle or JSON.

Closes #77.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/78)

### DIFF
--- a/test/test_variant.py
+++ b/test/test_variant.py
@@ -16,7 +16,10 @@
 Test simple properties of Variant objects, such as their trimming
 of shared prefix/suffix strings from ref/alt fields.
 """
-
+try:
+    import cPickle as pickle
+except ImportError:
+    import pickle
 from varcode import Variant
 from nose.tools import eq_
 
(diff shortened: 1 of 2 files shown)


varcode/variant.py
```diff
@@ -101,8 +109,8 @@ def __init__(self,
         # normalize the variant by trimming any shared prefix or suffix
         # between ref and alt nucleotide sequences and then
         # offset the variant position in a strand-dependent manner
-        trimmed_ref, trimmed_alt, prefix, suffix = trim_shared_flanking_strings(
-            self.original_ref, self.original_alt)
+        (trimmed_ref, trimmed_alt, prefix, suffix) = (
+            trim_shared_flanking_strings(self.original_ref, self.original_alt))
```
iskandr: Nitpick, but I don't think you need the extra parens around the function call, PEP8 seems pretty happy with arguments on the next line: http://legacy.python.org/dev/peps/pep-0008/#indentation

test/test_variant.py
```diff
@@ -98,3 +101,38 @@ def test_deletion_no_suffix():
     eq_(variant.start, 10)
     eq_(variant.end, 11)
     eq_(variant.short_description, "chr1 g.10_11delAA")
+
+def assert_variants_completely_equal(variant1, variant2):
+    eq_(variant1, variant2)
```
iskandr: Should something like `completely_equal` or `equal_with_info` be a member of `Variant`?

iskandr:
Other than a few questions, LGTM.

varcode/variant.py
timodonnell: Gotcha, the actual reason for that change was that the line was 80 chars, which my sublime pep8 plugin was complaining about (it wants <= 79 chars).

test/test_variant.py
timodonnell: I'd say let's wait to add that until we have a use case for it in non-test code

PR #79, opened 2015-04-30
Speed up PileupCollection.group_by_allele

timodonnell:
A loop in this function was accidentally running in time proportional to the total number of reads in the collection. This fix speeds up this function to an embarassing extent (over 1000X in my use case).

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/79)


iskandr:
LGTM and I'm really happy `del split[None]` is gone.

PR #82, opened 2015-05-05
Read and write json files

iskandr:
Added methods to `VariantCollection`:
* `variants.to_json_file(filename)`
* `VariantCollection.from_json_file(filename)`

Also fixed initialization of `end` property from basic fields.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/82)

### DIFF
--- a/varcode/variant.py
+++ b/varcode/variant.py
@@ -166,7 +166,7 @@ class Variant(object):
     # The identifying fields of a variant
     BasicFields = namedtuple(
         "BasicFields",
-        "contig start ref alt release")
+        "contig start end ref alt release")
 
     def fields(self):
         """
@@ -177,6 +177,7 @@ class Variant(object):
         return Variant.BasicFields(
             self.contig,
             self.start,
+            self.end,
             self.ref,
             self.alt,
             self.ensembl.release)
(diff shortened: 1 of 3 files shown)


varcode/variant.py
```diff
@@ -223,9 +223,13 @@ def __setstate__(self, fields):
         self.ensembl = cached_release(fields.pop("release"))
         self.info = fields.pop("info", {})
 
-        # Remaining fields are simple properties that just get set.
+        # Remaining fields  are simple properties that just get set.
         self.__dict__.update(fields)
 
+        # TODO: switch to interbase coordinates so that 'end' becomes a
+        # mandatory rather than derived field
+        self.end = self.start + max(0, len(self.ref) - 1)
```
timodonnell: This line may be right but it gives me pause. Can we add `end` to `fields()` instead? Then we would just extract it from the saved state here.

varcode/variant_collection.py
```diff
@@ -191,10 +191,18 @@ def to_json(self):
         """
         return json.dumps(self.__getstate__())
 
+    def to_json_file(self, filename):
```
timodonnell: Nit: I think `write_json` might be a better name for this function

timodonnell: or `write_json_file`, then the reverse can be called `read_json_file`

varcode/variant.py
iskandr: Seems reasonable, but one problem is that the `__init__` method doesn't take an `end` argument. Maybe I should make `end` optional and only compute it from the trimmed ref if it's absent?

timodonnell:
LGTM

PR #85, opened 2015-05-23
Use find_packages correctly

### LINKED ISSUE #84: pip installing Varcode doesn't seem to work lately
In my `site-packages`, I don't end up with `varcode` even though I do end up with `varcode-0.3.9.dist-info`. `pip list` thinks I have `varcode`, but `import varcode` fails.

tavinathanson:
Fixes #84 

`find_packages` appears to either take an exclusion list or no args, rather than just `"test"`. With no args, it does include both `varcode` and `test`, as evidenced by the `top_level.txt`:

```
test
varcode
```

Before this fix, `top_level.txt` is empty and `import varcode` doesn't work.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/85)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,8 +39,8 @@ except Exception as e:
 if __name__ == '__main__':
     setup(
         name='varcode',
-        packages=find_packages("test"),
-        version="0.3.10",
+        packages=find_packages(),
+        version="0.3.11",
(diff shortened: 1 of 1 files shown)


hammer:
Works for me!

Issue #40, opened 2015-03-24
Switch to interbase coordinates


iskandr:
Currently have a lot of special case logic for insertions (e.g. `if len(ref) == 0`), can use unified logic if the `start`/`end` coordinates are *between* bases. 

See: http://alternateallele.blogspot.de/2012/03/genome-coordinate-conventions.html

timodonnell:
This is a pile of annoying work, but I think we should either prioritize it or drop it. Having the read_evidence module using interbase coordinates while the rest of varcode uses inclusive is really untenable -- I'm hitting endless off by one errors and don't feel like I can write a blog post on read evidence pie charts when the code to generate them is constantly switching between two coordinate systems mandated by the same software package, all of which would change when we transition.

If we do change to interbase, have to figure out: what coordinate system should varcode use when reporting results to users (e.g. `str(variant)`). It will cause some confusion to give them 0-coordinates there I think (I accidentally sent someone 0-based coordinates to do PCR validation and confusion ensued) but on the other hand having it give different results than the variant properties is also confusing.

If there's a reasonable way to stick with 1-based coordinates, I wouldn't be opposed to that. Then I would just change read_evidence to use 1-based. And we should switch `Variant` instances to somehow use a `Locus` class (either have a locus or inherit from it) so we can work with loci and variants in a unified way.

hammer:
@iskandr @timodonnell have y'all had a chance to discuss this issue in person? Given the SciPy presentation last week and upcoming blog post this week, we should at least have a plan to communicate to potential users.

PR #87, opened 2015-05-29
Support for regular varcode variant instances in read evidence module

timodonnell:
The read evidence module has been in an awkward place since it uses a different `Variant` class than the rest of varcode, as it uses interbase coordinates whereas the rest of varcode uses 1-based coordinates. (Varcode will eventually migrate entirely to interbase; see #40). This change makes this situation mildly less annoying by having the read evidence module take care of converting regular varcode `Variant` instances to the needed interase format.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/87)

### DIFF
--- a/varcode/read_evidence/pileup_collection.py
+++ b/varcode/read_evidence/pileup_collection.py
@@ -23,7 +23,6 @@ import pandas
 import pysam
 import typechecks
 import pyensembl
-import typechecks
 
 from .. import Locus
 from . import Pileup, PileupElement, alignment_key, read_key
@@ -59,6 +58,7 @@ class PileupCollection(object):
         Raises a KeyError if this PileupCollection does not have a Pileup at
         the specified locus.
         '''
+        locus = to_locus(locus)
         if len(locus.positions) != 1:
             raise ValueError("Not a single-base locus: %s" % locus)
         return self.pileups[locus]
@@ -68,6 +68,7 @@ class PileupCollection(object):
         Return a new PileupCollection instance including only pileups for 
         the specified loci.
         '''
+        loci = [to_locus(obj) for obj in loci]
         single_position_loci = []
         for locus in loci:
             for position in locus.positions:
@@ -285,6 +286,7 @@ class PileupCollection(object):
         PileupCollection instances of the alignments that support that allele.
 
         '''
+        locus = to_locus(locus)
         read_to_allele = None
         loci = []
         if locus.positions:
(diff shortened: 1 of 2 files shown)


timodonnell:
Mind taking a quick perusal of this, @tavinathanson ?

varcode/read_evidence/pileup_collection.py
```diff
@@ -585,3 +591,22 @@ def from_bam(pysam_samfile, loci):
...
+    Since the read evidence module uses a different Variant class than the rest
+    of varcode, this is necessary. This should be removed once varcode switches
+    to interbase coordinates.
+    """
+    if isinstance(variant_or_locus, Locus):
+        return variant_or_locus
+    try:
+        return variant_or_locus.locus
+    except AttributeError:
+        # IMPORTANT: if varcode someday changes from inclusive to interbase
```
tavinathanson: Nit: maybe mark this with `FIXME` or `TODO` for easy discovery of FIXMEs/TODOs (I should probably put this suggestion in the wiki)

tavinathanson:
@timodonnell I'm somewhat confused by the `read_evidence` module; for starters, I don't see your version of `Variant` defined anywhere, or for that matter, code that actually uses the `read_evidence` module.

Update: Ah, I'm now looking at https://github.com/hammerlab/ovarian-cancer/blob/53c92e7ddc0b81b0864c7d28fd112c75c8747472/projects/pt189/read_evidence_plot.py to get a feel for usage.

varcode/read_evidence/pileup_collection.py
```diff
@@ -585,3 +591,22 @@ def from_bam(pysam_samfile, loci):
             if close_on_completion:
                 pysam_samfile.close()
 
+def to_locus(variant_or_locus):
```
tavinathanson: I'm confused about why `variant` and `locus` are combined here?

(In general, I feel like `x_or_y` isn't good practice, but I don't yet understand why the code wants this in the first place.)

varcode/read_evidence/pileup_collection.py
```diff
@@ -585,3 +591,22 @@ def from_bam(pysam_samfile, loci):
...
+    """
+    Return a Locus object for a Variant instance.
+
+    Since the read evidence module uses a different Variant class than the rest
```
tavinathanson: Maybe mention where someone can go to see the definition of this different Variant class?

timodonnell: Thanks @tavinathanson , good call. I'm going to improve the docs of this function to better explain what's going on and then merge. This code is pretty awful right now but should get hopefully much simpler once all of varcode changes to interbase coordinates.

PR #93, opened 2015-06-18
Support loading VCFs from URL

### LINKED ISSUE #91: support loading VCFs over HTTP
It would be useful if the `load_vcf` function accepted URLs.

Current use case for this is opening VCFs on my laptop that are on demeter hdfs. It would be convenient to use HTTP for that.

timodonnell:
Add support for loading VCFs over any protocol supported by `urlopen` (HTTP, HTTPS, FTP). Closes #91.

Also adds a test for loading gzip'd VCFs from files and over HTTP.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/93)


timodonnell:
I'm actually going to do a new PR with a much faster implementation

PR #102, opened 2015-07-06
Manually set compression in read_vcf_into_dataframe

### LINKED ISSUE #101: varcode.load_vcf_fast used 0.16.1 Pandas options
`varcode.load_vcf_fast` uses `compression='infer'` which is available in Pandas 0.16.1, but not earlier.
https://github.com/hammerlab/varcode/blob/master/varcode/vcf.py#L368 

Should we upgrade the Pandas version needed or change the compression option?
See: http://pandas.pydata.org//pandas-docs/version/0.16.1/generated/pandas.read_csv.html

timodonnell:
Closes #101

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/102)

### DIFF
--- a/varcode/vcf.py
+++ b/varcode/vcf.py
@@ -363,9 +363,15 @@ def read_vcf_into_dataframe(path, include_info=False, chunk_size=None):
     else:
         raise NotImplementedError("Only local files are supported.")
 
+    compression = None
+    if path.endswith(".gz"):
+        compression = "gzip"
+    elif path.endswith(".bz2"):
+        compression = "bzip2"
+
     reader = pandas.read_table(
         path,
-        compression='infer',
+        compression=compression,
         comment="#",
         chunksize=chunk_size,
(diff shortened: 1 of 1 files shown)


varcode/vcf.py
```diff
@@ -363,9 +363,15 @@ def read_vcf_into_dataframe(path, include_info=False, chunk_size=None):
     else:
         raise NotImplementedError("Only local files are supported.")
 
+    compression = None
+    if path.endswith(".gz"):
+        compression = "gzip"
+    elif path.endswith(".bz2"):
+        compression = "bzip2"
```
arahuja: I think we want `bz2` here, 

From the Pandas doc: `compression : {‘gzip’, ‘bz2’, None}, default None`

timodonnell:
Good catch -- updated

arahuja:
Cool, LGTM

PR #106, opened 2015-07-20
Allow Varcode to work with mouse data via Genome

tavinathanson:
This is a quick fix to allow Varcode to work with mouse data (and an initial test for that) by accepting `Genome` objects.

It's certainly confusing to be using `ensembl` and `genome` somewhat interchangeably here. I'd like a follow-up PR (and will create an issue, if this order-of-things sounds good to you) to liberate Varcode from Ensembl much like https://github.com/hammerlab/pyensembl/pull/99, but for now this seems useful.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/106)


timodonnell:
Many thanks for doing this in a way that does not break all of our existing code (which passes in `ensembl_version` to `load_vcf`)!

tavinathanson:
@timodonnell Fewer tests for me to fix, too! The intent of https://github.com/hammerlab/pyensembl/pull/99 is also to keep the `EnsemblRelease` API intact despite refactoring the innards and adding a superclass to it.

iskandr:
Finding lots of little bugs in the PyEnsembl PR we merged, but also finding a design flaw wherein each GenomeSource only tells you the command you need to download e.g. a protein FASTA but that command can't be run in isolation (since the pyensembl script always requires a GTF). 

I guess we'll figure out a better design when you come back.

iskandr:
To clarify on the last comment, to install the mouse genome you need to run:

```
pyensembl install --reference-name "GRCm38" --transcript-fasta-path-or-url "ftp://ftp.ensembl.org/pub/release-81/fasta/mus_musculus/cdna/Mus_musculus.GRCm38.cdna.all.fa.gz" --protein-fasta-path-or-url "ftp://ftp.ensembl.org/pub/release-81/fasta/mus_musculus/pep/Mus_musculus.GRCm38.pep.all.fa.gz" --gtf-path-or-url "ftp://ftp.ensembl.org/pub/release-81/gtf/mus_musculus/Mus_musculus.GRCm38.81.gtf.gz"
```

but if this genome is missing you'll see errors like:

```
nose.proxy.ValueError: Genome sequence data (GenomeSource(transcript_fasta_path_or_url=ftp://ftp.ensembl.org/pub/release-81/fasta/mus_musculus/cdna/Mus_musculus.GRCm38.cdna.all.fa.gz)) is not currently installed for this genome source. Run pyensembl install --transcript_fasta_path_or_url "ftp://ftp.ensembl.org/pub/release-81/fasta/mus_musculus/cdna/Mus_musculus.GRCm38.cdna.all.fa.gz" or call Genome(reference_name="GRCm38", transcript_fasta_path_or_url="ftp://ftp.ensembl.org/pub/release-81/fasta/mus_musculus/cdna/Mus_musculus.GRCm38.cdna.all.fa.gz")).install()
```

iskandr:
Additionally, it looks like EnsemblReleaseSource no longer gets used and thus even human Ensembl references emit an error like the one above.

tavinathanson:
Offline discussion re the design flaw: it still "works", but might be an annoyance to run multiple commands. We agreed to table that until https://github.com/hammerlab/pyensembl/issues/100 is addressed.

tavinathanson:
Re the bugs: @iskandr and I worked through them together and they're summarized/addressed here: https://github.com/hammerlab/pyensembl/pull/108

Offline discussion re `EnsemblReleaseSource`: that was intended to be used, and is used in the PR above.

tavinathanson:
@iskandr I don't remember how we left it with this PR, since it's hacky but not entirely awful. What would you like to see before we merge?

iskandr:
@tavinathanson Let's merge this, and get mouse epitope predictions out of Topiary, and figure out a better organization next week.

PR #112, opened 2015-07-29
fix logic for determining whether the protein length decreases

### LINKED ISSUE #111: PrematureStop called as a Deletion
Here's an example of a variant that should be a PrematureStop but varcode calls a Deletion (SnpEff calls stop_gained):

Variant(contig=7, start=117144357, ref="C", alt="A", ensembl=75)

leekaiinthesky:
Fixes #111

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/112)


iskandr:
:+1:  Thanks Lee-kai.

Issue #105, opened 2015-07-20
Double mutations in a MAF file cause error


kippakers:
Scanning my logs, a good chunk of my failed jobs have something along these lines with a double mutation having an off-by-one location end:

```python
ValueError: Expected variant 1:109461324 'GG' > 'TT' to end at 109461325 but got 109461326
```
This appears to be arising from varcode (I'm running topiary):
```python
INFO:root:Building MHC binding prediction type for alleles ['HLA-A*30:01', 'HLA-A*02:01', 'HLA-B*38:01', 'HLA-B*48:01', 'HLA-C*08:03', 'HLA-C*12:03'] and epitope lengths [9]
INFO:root:netMHCcons finished with return code 0
INFO:root:netMHCcons took 0.0690 seconds
Traceback (most recent call last):
  File "/hpc/users/akersn01/.local/bin/topiary", line 5, in <module>
    pkg_resources.run_script('topiary==0.0.6', 'topiary')
  File "/hpc/packages/minerva-common/py_packages/2.7/lib/python2.7/site-packages/distribute-0.6.10-py2.7.egg/pkg_resources.py", line 461, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/hpc/packages/minerva-common/py_packages/2.7/lib/python2.7/site-packages/distribute-0.6.10-py2.7.egg/pkg_resources.py", line 1194, in run_script
    execfile(script_filename, namespace, namespace)
  File "/hpc/users/akersn01/.local/lib/python2.7/site-packages/topiary-0.0.6-py2.7.egg/EGG-INFO/scripts/topiary", line 99, in <module>
    main()
  File "/hpc/users/akersn01/.local/lib/python2.7/site-packages/topiary-0.0.6-py2.7.egg/EGG-INFO/scripts/topiary", line 59, in main
    variants = variant_collection_from_args(args)
  File "/hpc/users/akersn01/.local/lib/python2.7/site-packages/topiary-0.0.6-py2.7.egg/topiary/args.py", line 61, in variant_collection_from_args
    variant_collections.append(varcode.load_maf(maf_path))
  File "/hpc/users/akersn01/.local/lib/python2.7/site-packages/varcode-0.3.12-py2.7.egg/varcode/maf.py", line 163, in load_maf
    end_pos))
```
here is the offending line of the maf:
```
[...]

PR #115, opened 2015-08-02
Warn when variants in MAF file have wrong end position

iskandr:
Some MAFs in the wild seem to have end positions which violate what's expected with base-1 inclusive coordinates, so warn about unexpected end positions instead of raising an exception. (fixes https://github.com/hammerlab/varcode/issues/105)

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/115)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ if __name__ == '__main__':
     setup(
         name='varcode',
         packages=find_packages(),
-        version="0.3.15",
+        version="0.3.16",
         description="Variant annotation in Python",
         long_description=readme,
         url="https://github.com/hammerlab/varcode",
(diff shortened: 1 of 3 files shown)


tavinathanson:
Hmm. Do we really *never* want to `ValueError` here? What if we warn only in a specific off-by-one case, and `ValueError` otherwise?

iskandr:
@tavinathanson I thought about trying to figure out what a safe category of errors is, but...what if our normalization of the ref/alt sequences makes the "safe" error range greater than 1?

tavinathanson:
Thinking more about this: I feel like this still is saying more about the wonkiness of the `MAF` file than anything else, and the user probably should know that their `MAF` file is weird.

What about `ValueError` by default, unless the user loads the `MAF` as follows: `variants = varcode.load_maf("tcga-ovarian-cancer-variants.maf", validate_maf=False)`?

tavinathanson:
Ah, but that would need to be annoyingly passed through from `topiary`, given Kipp's use case.

Eh. Bleh. Feel free to just merge.

PR #117, opened 2015-08-13
Fix #116 and call PrematureStop when stop codon is added in the middle of an insertion

### LINKED ISSUE #116: PrematureStop called as Silent
E.g.

    #CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	SAMPLE
    7	117182143	.	T	TGTA	.	.	.	GT	0/1

This changes GTA to GTGTAA which introduces a premature stop.

In in_frame_coding_effect.py line 290, we only return a PrematureStop if the length of the protein subsequence decreases. However, here we compare GTA -> V to GTGTAA -> V(*), and we erroneously decide they're the same.

In order to really be sure the protein length is decreasing, we'd need to bring in the next codon as well when comparing the subsequences. (If the next codon happens to be a stop codon in both the reference and the mutant, then I suppose calling the variant Silent is correct. But by this logic even a long insertion including a stop codon could theoretically be Silent, and we don't check for that.)

Another approach would be to include stop codons (*) in the translation. That seems to be part of a bigger design decision though, so defer to @iskandr's preference.

leekaiinthesky:
Fixes #116. Also ran into an issue that if a stop codon is not at the end of an in-frame coding variant, it will not get called a PrematureStop. I seems like it still should though, so I changed that too.

Here are three examples that were not called as PrematureStop before but are now.

Called Silent but should be PrematureStop
`7	117182143	.	T	TGTA	.	.	.	GT	0/1`
Called Deletion but should be PrematureStop
`7	117182143	.	TAACAG	TATAAC	.	.	.	GT	0/1`
Called Insertion but should be PrematureStop
`7	117182143	.	TA	TATAGGTT	.	.	.	GT	0/1`

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/117)

### DIFF
--- a/varcode/in_frame_coding_effect.py
+++ b/varcode/in_frame_coding_effect.py
@@ -177,9 +177,10 @@ def in_frame_coding_effect(
                 aa_alt=inserted_amino_acids)
         else:
             # inserting inside a reference codon
+            # include an extra codon at the end of the reference so that if we insert a stop before a stop, we can return Silent
             ref_codon = sequence_from_start_codon[
-                first_ref_codon_index * 3:first_ref_codon_index * 3 + 3]
-            last_ref_codon_index = first_ref_codon_index
+                first_ref_codon_index * 3:first_ref_codon_index * 3 + 6]
+            last_ref_codon_index = first_ref_codon_index + 1
             # split the reference codon into nucleotides before/after insertion
             prefix = ref_codon[:offset_in_first_ref_codon + 1]
             suffix = ref_codon[offset_in_first_ref_codon + 1:]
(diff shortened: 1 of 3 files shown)


varcode/in_frame_coding_effect.py
```diff
@@ -162,7 +162,7 @@ def in_frame_coding_effect(
             # then we can just translate the inserted sequence since it's on
             # codon boundary
             inserted_amino_acids = translate(alt, first_codon_is_start=False)
-            if alt[-3:] in STOP_CODONS:
+            if STOP_CODONS.intersection(alt[3*i:3*i + 3] for i in xrange(len(alt)/3)):
```
iskandr: Sadly, we lost `xrange` in Py3, do you mind changing this to `six.moves.range`?

iskandr:
Thanks for catching this. Aside from making sure it works with Python 3, do you mind adding the example variants to `test_effect_classes.py`?

leekaiinthesky:
Ok, added! Thanks for the catch with compatibility.

PR #119, opened 2015-08-28
Minor problem in Variant.__init__

iskandr:
The "fast" path for initializing a Variant object didn't convert the `start` parameter to an int. Also added support for creating the variant's associated genome from the string name of a reference.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/119)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ if __name__ == '__main__':
     setup(
         name='varcode',
         packages=find_packages(),
-        version="0.3.17",
+        version="0.3.18",
         description="Variant annotation in Python",
         long_description=readme,
         url="https://github.com/hammerlab/varcode",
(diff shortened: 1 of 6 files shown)


tavinathanson:
If we're letting a `Genome` be an `int` `release` or a `str` `reference_name`, we might as well do the same in `load_vcf` so that @timodonnell's `release_num` code doesn't break?

Offline discussion: the `load_vcf` `genome` argument can just be passed through to `Variant`?

tavinathanson:
Only question I have is: seems like `infer_genome` is basically duplicated logic? https://github.com/hammerlab/varcode/blob/fix-variant-start-and-string-reference/varcode/variant.py#L114 vs. https://github.com/hammerlab/varcode/blob/fix-variant-start-and-string-reference/varcode/vcf.py#L440?

PR #125, opened 2015-09-28
Remove read_evidence and locus modules

### LINKED ISSUE #124: move read_evidence module and Locus class to varlens
varcode is cleaner if the functionality for reading BAM files is in another library, and I think the library I'm developing called `varlens` is a reasonable place for that (https://github.com/timodonnell/varlens -- soon to be in hammerlab once it's cleaned up a bit). It will also hopefully lead to a less convoluted implementation of `read_evidence`.

This will also mitigate the absurd situation where varcode uses 1-based inclusive coordinates but `varcode.read_evidence` uses interbase coordinates.

timodonnell:
This functionality has been moved to the `varlens` repository. Closes #124

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/125)


iskandr:
Missing a change to the package version?

iskandr:
Otherwise LGTM

timodonnell:
Bumped version and also removed pysam dependency

Issue #129, opened 2015-10-20
Intragenic variants do not have a short_description field


armish:
See https://github.com/hammerlab/cycledash/issues/855 and https://github.com/hammerlab/cycledash/pull/857 for context.

Suggestion from @iskandr:
> Everything should have a short_description, I should probably provide a default value which is something like `self.__class__.__name__.lower()`

PR #130, opened 2015-10-20
All effects should have a default `short_description` field

armish:
See #129.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/130)

### DIFF
--- a/varcode/effects.py
+++ b/varcode/effects.py
@@ -40,9 +40,12 @@ class MutationEffect(object):
 
     @property
     def short_description(self):
-        raise ValueError(
-            "Property short_description not implemented for %s" % (
-                self.__class__.__name__,))
+        """
+        A short but human-readable description of the effect.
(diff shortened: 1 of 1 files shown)


varcode/effects.py
```diff
@@ -115,6 +118,7 @@ class Intragenic(MutationEffect):
     apparently does happen sometimes, maybe some genes have two distinct sets
     of exons which are never simultaneously expressed?
     """
+    short_description = "intragenic"
```
arahuja: Is this almost always just the class name?

armish: not always; coding effects usually give you the scientific notation of a variant. Example from https://github.com/hammerlab/varcode/blob/master/examples/varcode-quick_start.ipynb:

```
> myVariant.short_description
'chr7 g.140453136A>T'
```

iskandr:
LGTM

Issue #136, opened 2016-02-23
ExonicSpliceSite mutations are classified as Noncoding


arahuja:
`ExonicSpliceSite` is classified as noncoding. It is an instance of `Exonic` and `SpliceSite` but not of `NonsilentCodingMutation` even though it can have an alternate effect of `Substitution`

Maybe not an error, but wanted to point it out. I think this variant type should carry through when using `drop_silent_and_noncoding` since it can result in a substitution?

```
In [16]: from varcode import Variant

In [17]: v = Variant('17', 67190529, ref='C', alt='T', ensembl=ensembl)

In [18]: v.effects()
Out[18]:
<EffectCollection with 5 elements>
  -- ExonicSpliceSite(exon=Exon(exon_id=ENSE00003661242, gene_name=ABCA10, contig=17, start=67190526, end=67190636), alternate_effect=Substitution(variant=chr17 g.67190529C>T, transcript_name=ABCA10-001, transcript_id=ENST00000269081, effect_description=p.E448K))
  -- ExonicSpliceSite(exon=Exon(exon_id=ENSE00003661242, gene_name=ABCA10, contig=17, start=67190526, end=67190636), alternate_effect=Substitution(variant=chr17 g.67190529C>T, transcript_name=ABCA10-201, transcript_id=ENST00000416101, effect_description=p.E448K))
  -- ThreePrimeUTR(variant=chr17 g.67190529C>T, transcript_name=ABCA10-004, transcript_id=ENST00000518929)
  -- ThreePrimeUTR(variant=chr17 g.67190529C>T, transcript_name=ABCA10-003, transcript_id=ENST00000522406)
  -- ExonicSpliceSite(exon=Exon(exon_id=ENSE00003661242, gene_name=ABCA10, contig=17, start=67190526, end=67190636), alternate_effect=Substitution(variant=chr17 g.67190529C>T, transcript_name=ABCA10-005, transcript_id=ENST00000523419, effect_description=p.E448K))

In [19]: v.effects().drop_silent_and_noncoding()
Out[19]: <EffectCollection with 0 elements>
```

iskandr:
Huh, this is tricky since whether a `ExonicSpliceSite` splice site mutation counts as non-synonymous also depends on what the alternate effect is. It could easily be `Silent`, no? 

I'm going to count it as a bug for `drop_silent_and_noncoding()` to drop the variants in your example.

iskandr:
[...]

PR #137, opened 2016-02-23
Modest change to filtering of coding mutations include ExonicSpliceSite 

iskandr:
Aside from change to use `effect.modifies_protein_sequence` and updating versions of Ensembl used by tests, the rest of the changes are littered with a bunch of PEP8 formatting fixes.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/137)


varcode/util.py
```diff
@@ -48,45 +48,52 @@ def random_variants(
 
     variants = []
 
-    while len(variants) < count:
-        transcript_id = rng.choice(transcript_ids)
-        transcript = ensembl.transcript_by_id(transcript_id)
+    # we should finish way before this loop is over but just in case
+    # something is wrong with PyEnsembl we want to avoid an infinite loop
```
tavinathanson: What sort of infinite loop are we avoiding here?

varcode/effect_ordering.py
```diff
@@ -12,13 +12,38 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .effects import *
+from .effects import (
```
tavinathanson: Why the change?

varcode/effect_collection.py
```diff
@@ -143,8 +143,7 @@ def drop_silent_and_noncoding(self):
         """
         Create a new EffectCollection containing only non-silent coding effects
         """
-        return self.filter(
-            lambda effect: isinstance(effect, NonsilentCodingMutation))
+        return self.filter(lambda effect: effect.modifies_protein_sequence)
```
tavinathanson: I think this makes sense?

varcode/util.py
iskandr: The one that occurred to me when Ensembl 83 had bad identifier names for transcripts so none of Transcript objects had valid `sequence` fields.

varcode/effect_ordering.py
iskandr: I'm always jumpy about `import *` but no new reason.

varcode/effect_collection.py
iskandr: That property defaults to False on the base class `MutationEffect` and gets overridden as True on `NonsilentCodingMutation`. The main change in this PR is adding a dynamic property to `ExonicSpliceSite` which checks its alternate effect  -- before that change `effect.modifies_protein_sequence` and `isinstance(effect, NonsilentCodingMutation)` meant the same thing.

varcode/util.py
```diff
@@ -48,45 +48,52 @@ def random_variants(
...
+            transcript_offset = transcript.spliced_offset(base1_genomic_position)
 
-        try:
-            seq = transcript.sequence
-        except ValueError as e:
-            logging.warn(e)
-            # can't get sequence for non-coding transcripts
-            continue
+            try:
+                seq = transcript.sequence
```
tavinathanson: As discussed offline, this shouldn't catch since it doesn't throw anymore

varcode/effect_ordering.py
tavinathanson: :+1:

varcode/effect_collection.py
tavinathanson: Yeah, that was my understanding of the changes as well; just wondering if there are classes that need to be modified, etc.

varcode/util.py
tavinathanson: :+1:

tavinathanson:
LGTM

iskandr:
Fixes https://github.com/hammerlab/varcode/issues/136

arahuja:
@iskandr do you want to add `modifies_coding_sequence` to `ExonicSpliceSite` as well, also implemented through the alternate effect?

iskandr:
@arahuja Good idea!

arahuja:
@iskandr cool, LGTM

PR #138, opened 2016-03-04
added is_deletion, is_insertion, and is_indel properties to variants

iskandr:
The other changes are minor PEP8 fixes and updating the copyright year.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/138)

### DIFF
--- a/test/test_variant.py
+++ b/test/test_variant.py
@@ -44,6 +44,7 @@ def test_insertion_shared_prefix():
     assert variant.is_indel
     assert variant.is_insertion
     assert not variant.is_deletion
+    assert not variant.preserves_reading_frame
 
 def test_insertion_no_prefix():
     variant = Variant(1, start=11, ref="", alt="T")
@@ -59,6 +60,7 @@ def test_insertion_no_prefix():
     assert variant.is_indel
     assert variant.is_insertion
     assert not variant.is_deletion
+    assert not variant.preserves_reading_frame
 
 def test_substitution_no_prefix():
     variant = Variant(1, start=11, ref="A", alt="T")
@@ -74,6 +76,7 @@ def test_substitution_no_prefix():
     assert not variant.is_indel
     assert not variant.is_insertion
     assert not variant.is_deletion
+    assert variant.preserves_reading_frame
 
 
 def test_substitution_shared_prefix():
@@ -90,6 +93,7 @@ def test_substitution_shared_prefix():
     assert not variant.is_indel
     assert not variant.is_insertion
     assert not variant.is_deletion
+    assert variant.preserves_reading_frame
 
 
 def test_deletion_shared_suffix():
@@ -106,6 +110,7 @@ def test_deletion_shared_suffix():
     assert variant.is_indel
     assert not variant.is_insertion
     assert variant.is_deletion
+    assert not variant.preserves_reading_frame
 
 
 def test_deletion_no_suffix():
@@ -122,6 +127,7 @@ def test_deletion_no_suffix():
     assert variant.is_indel
     assert not variant.is_insertion
     assert variant.is_deletion
+    assert not variant.preserves_reading_frame
 
 def test_serialization():
     variants = [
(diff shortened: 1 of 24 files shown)


varcode/variant_collection.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2015. Mount Sinai School of Medicine
```
tavinathanson: You sure it's 2016? Whoa.

test/test_variant.py
```diff
@@ -41,6 +41,9 @@ def test_insertion_shared_prefix():
     eq_(variant.start, 11)
     eq_(variant.end, 11)
     eq_(variant.short_description, "chr1 g.11_12insT")
+    assert variant.is_indel
```
tavinathanson: Nit, but: `ok_` might be more `nose` idiomatic?

varcode/variant.py
```diff
@@ -681,3 +681,24 @@ def _exonic_transcript_effect(self, exon, exon_number, transcript):
...
                 alternate_effect=coding_effect_annotation)
         return coding_effect_annotation
+
+    @property
+    def is_insertion(self):
+        """
+        Does this variant represent the insertion of nucleotides into the
+        reference genome?
+        """
+        return (len(self.ref) < len(self.alt)) and self.alt.startswith(self.ref)
```
tavinathanson: `self.alt.startswith(self.ref)` seems like a confusing thing to do when dealing with `""`? Am I misunderstanding?

tavinathanson: Would `len(self.ref) == 0` not do all of the above?

test/test_variant.py
iskandr: I like eq_ because it comes with a default message of the wrong value but with booleans, what does `ok_` buy us?

varcode/variant.py
iskandr: I tried to write it to not rely on any kind of normalization, so that `CT`>`C` would be recognized as a deletion and `C`>`CT` would be recognized as an insertion.

tavinathanson: Doesn't the code specifically do normalization before we get here?

test/test_variant.py
tavinathanson: Hence being a nit; http://nose.readthedocs.org/en/latest/testing_tools.html#nose.tools.ok_ "Saves 3 whole characters!"

Completely up to you.

varcode/variant_collection.py
iskandr: Copyright (c) 2157. Mount Sinai School of Medicine 

(the Python3 vs Python4 schism will be mostly healed by then, with >20% Python4 adoption)

test/test_variant.py
tavinathanson: (Technically, `ok_` also prevents tests from being optimized away with `python -O`, but who cares? https://github.com/nose-devs/nose/issues/504)

varcode/variant.py
iskandr: Yeah, it's done in the constructor but that's always felt a little controversial (since we're changing the representation away from how it's commonly displayed or what you see in VCFs) and I wanted to limit reliance on that normalization just in the variant effect prediction (where it matters the most). 

However, if you think that's more confusing then I can just simply the checks here and add comments saying we're assuming the variants have been normalized.

tavinathanson: Happy with either option as long the assumption is clearly commented.

varcode/variant.py
```diff
@@ -681,3 +681,41 @@ def _exonic_transcript_effect(self, exon, exon_number, transcript):
...
+        # nucleotides are an empty string.
+        return (len(self.ref) > len(self.alt)) and self.ref.startswith(self.alt)
+
+    @property
+    def is_indel(self):
+        """Is this variant either an insertion or deletion?"""
+        return self.is_insertion or self.is_deletion
+
+    @property
+    def preserves_reading_frame(self):
```
arahuja: This might be misleading if it's a stoploss?

iskandr: I couldn't think of a better name for "variants which preserve the reading frame if they fall in the coding sequence, excluding the start and stop codons". Ideas?

arahuja: Is there an example use for that?

iskandr: If you already know that the variant falls in the coding sequence and are trying to determine the impact of the variant.

PR #139, opened 2016-03-14
Quotes around nucleotides in Variant representation

iskandr:
Instead of representing empty strings with `.`, instead using single quotes around all string fields of a `Variant`. This brings us closer to having a string representation of a variant which can be evaluated by Python to get back to the same object (the reference name isn't quite there). 

Also, got rid of `Variant.preserves_reading_frame` since I've come around to @arahuja's opinion that this is a confusing property.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/hammerlab/varcode/139)

### DIFF
--- a/varcode/variant.py
+++ b/varcode/variant.py
@@ -711,11 +711,3 @@ class Variant(object):
     def is_indel(self):
         """Is this variant either an insertion or deletion?"""
         return self.is_insertion or self.is_deletion
-
-    @property
-    def preserves_reading_frame(self):
-        """
-        If this variant occurs in the coding region of a gene, would it
-        prerseve the ORF or create a frameshift mutation?
-        """
-        return (len(self.alt) - len(self.ref)) % 3 == 0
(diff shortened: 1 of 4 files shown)


varcode/variant.py
```diff
@@ -180,11 +180,11 @@ def reference_name(self):
         return self.ensembl.reference_name
 
     def __str__(self):
-        return "Variant(contig=%s, start=%d, ref=%s, alt=%s, reference_name=%s)" % (
+        return "Variant(contig='%s', start=%d, ref='%s', alt='%s', reference_name='%s')" % (
             self.contig,
             self.start,
-            self.ref if self.ref else ".",
-            self.alt if self.alt else ".",
+            self.ref,
```
arahuja: does this need to check for ref or alt being `None` or does that never happen?

arahuja:
LGTM, one question on whether the alt/ref should still be explicitly set as empty string

varcode/variant.py
iskandr: We check the length of both `alt` and `ref` in the initializer so I don't think it's possible for them to be `None` in a constructed `Variant`.

PR #140, opened 2016-03-29
Preserve contig name

iskandr:
Added option to preserve user-supplied contig name, so we don't always assume we should be e.g. capitalizing the chromosome and trimming the 'chr' prefix.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/140)


tavinathanson:
LGTM, but the naming is a little wonky IMO. What about `normalize_contigs=True` by default and optionally `False`?

iskandr:
@tavinathanson I wasn't sure what to name it since the most obvious name, `normalize_contig_name` was already taken by the imported function. I guess `normalize_contigs` is OK but I'm a little confused by the plural `s` at the end.

iskandr:
Ah, just realized the normalization function is called `normalize_chromosome`, so going with `normalize_contig_name` as the parameter.

PR #145, opened 2016-04-22
Parse and expose sample info, including for multisample VCFs

### LINKED ISSUE #30: Attach genotypes and other sample information to Variants

### LINKED ISSUE #96: Is a VariantCollection single sample or multiple samples?
If it's multiple samples - where's the best place to save the SampleId - right now unless I set `distinct=True` it would throw out repeated varaints?

timodonnell:
Add per-sample info to vcf loading. Works the same whether you use `load_vcf`
or `load_vcf_fast`.

The VariantCollection.metadata entry for each variant now includes a
`sample_info` entry, which maps samples names to a dict of pyvcf-parsed
per-sample attributes for that sample.

I think this takes care of the "support multisample vcf" issues. Summary is:

 * a `VariantCollection` is multisample: it contains whatever variants were in your VCF
 * the `VariantCollection.metadata` dict includes per-sample info (as well as across-sample info from the INFO column and filters, etc.)
 * if you care about what sample a variant was "called" in, it is left to the user to look at the genotype and decide what to do. For most VCFs this will be available in `VariantCollection.metadata[variant]['sample_info']['GT']` and the value '0/0' means no-call and anything else means there was a call. We don't assume anything about this field here though (similarly to pyvcf).

Closes #96 . Closes #30 .

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/145)

### DIFF
--- a/test/test_vcf.py
+++ b/test/test_vcf.py
@@ -78,7 +78,8 @@ def test_pandas_and_pyvcf_implementations_equivalent():
         {'path': data_path("multiallelic.vcf")},
         {'path': data_path("mutect-example.vcf")},
         {'path': data_path("strelka-example.vcf")},
-        {'path': data_path("mutect-example-headerless.vcf"), 'genome': cached_release(75)},
+        {'path': data_path("mutect-example-headerless.vcf"),
+            'genome': cached_release(75)},
     ]
     if RUN_TESTS_REQUIRING_INTERNET:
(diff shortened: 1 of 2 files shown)


iskandr:
It's a little confusing but that's probably unavoidable complexity. Otherwise LGTM.

Issue #142, opened 2016-04-18
Change meaning of ref and alt fields


iskandr:
Currently `variant.ref` and `variant.alt` don't contain any shared prefix or suffix. This confuses user since they expect whatever values passed in for these fields to be unchanged. 

Breaking change: rename the current `ref` and `alt` fields to `trimmed_ref` and `trimmed_alt` and get rid of `original_ref` and `original_alt`.

PR #146, opened 2016-04-25
Fix variant pickling

tavinathanson:
Pickle the original `ref`/`alt`/`start` vs. the normalized values. This allows for all to be present in the reconstituted object.

This is related to, but shouldn't need to wait for, https://github.com/hammerlab/varcode/issues/142

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/146)

### DIFF
--- a/varcode/variant.py
+++ b/varcode/variant.py
@@ -234,6 +234,9 @@ class Variant(object):
             self.end,
             self.ref,
             self.alt,
+            self.original_start,
+            self.original_ref,
+            self.original_alt,
             self.ensembl.release)
 
     def __eq__(self, other):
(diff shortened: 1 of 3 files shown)


varcode/variant.py
```diff
@@ -248,18 +251,18 @@ def __eq__(self, other):
...
+        # When pickle.load'ing, we want the original values vs. the normalized values.
+        # Normalization will happen upon object creation.
+        return [self.contig, self.original_start, self.original_ref, self.original_alt, self.ensembl.release]
 
     def __setstate__(self, fields):
-        # This field require special logic.
-        self.ensembl = cached_release(fields.pop("release"))
+        # Special logic needed for JSON serialization, since EnsemblRelease
+        # is not JSON serializable.
+        # TODO: this does not work for Genome objects; only EnsemblRelease.
```
iskandr: How much more work would it be to get this working with arbitrary Genome objects?

varcode/variant.py
```diff
@@ -248,18 +251,18 @@ def __eq__(self, other):
...
+        ensembl = cached_release(fields.pop())
```
iskandr: This looks like it's always going to return a human reference (even if the original was mm10). Maybe worth adding a test with a mouse genome?

varcode/variant.py
tavinathanson: Not sure the best way to do that. I imagine we'd need to add JSON serialization for a lot of different classes. (Pickling already works with `Genome` objects.)

I figured I'd limit this PR to fixing one clear bug: original values weren't being reconstituted.

varcode/variant.py
tavinathanson: Agreed, that seems like a problem; mind if I create a separate issue since it's an old bug that isn't introduced here?

iskandr: Sure, go for it.

tavinathanson:
@iskandr Good to merge?

PR #149, opened 2016-05-11
Use versioneer to manage version number

arahuja:
I wanted to add this to this repository as right now master and pypi both have varcode==0.4.2 but different commits.  The last PR #148 is needed for some fixes but unreleased (as there was no version bump).

This uses versioneer to manage the version, and the current version (in this branch) would be 
```
>>> varcode.__version__
'v0.4.2+27.g165ef5f'
```

This also changes Travis to release on tags only, but a new tag can be added with
```
git tag <version>
git push --tags
```
In the `setup.cfg` I set `tag_prefix='v'` as for the one release now there is a `v` prefix.

In the second commit, I added a `CHANGELOG.md` generated automatically by `github_changelog_generator`.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/149)


iskandr:
LGTM

Issue #128, opened 2015-10-06
Support collection.as_dataframe()


tavinathanson:
I think we had that logic and it somehow disappeared?

PR #150, opened 2016-05-12
Add as_dataframe to EffectCollection

arahuja:
This adds an `as_dataframe` to EffectCollection - happy to add or rename any columns.

Also, a few helper properties to `Variant` that I have been using:
- `is_snv`
- `is_transition`
- `is_transversion`

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/150)

### DIFF
--- a/varcode/effect_collection.py
+++ b/varcode/effect_collection.py
@@ -14,6 +14,7 @@
 
 from __future__ import print_function, division, absolute_import
 from collections import Counter, OrderedDict
+import pandas as pd
 
 from typechecks import require_iterable_of
 
@@ -262,3 +263,26 @@ class EffectCollection(Collection):
         for effect in self:
             counter[effect.gene_name] += 1
         return counter
+
+    def to_dataframe(self):
+        """Build a dataframe from the effect collection"""
+        def row_from_effect(effect):
+            row = {}
+            row['chr'] = effect.variant.contig
+            row['position'] = effect.variant.start
+            row['ref'] = effect.variant.ref
+            row['alt'] = effect.variant.alt
+            row['gene_id'] = effect.gene_id
+            row['gene_name'] = effect.gene_name
+            row['transcript_id'] = effect.transcript_id
+            row['transcript_name'] = effect.transcript_name
+            row['variant'] = str(effect.variant)
+            row['is_snv'] = effect.variant.is_snv
+            row['is_indel'] = effect.variant.is_indel
+            row['is_transversion'] = effect.variant.is_transversion
+            row['is_transition'] = effect.variant.is_transition
+            row['effect'] = str(effect)
+            row['effect_type'] = effect.__class__.__name__
+            row['effect_description'] = effect.short_description
+            return row
+        return pd.DataFrame.from_records([row_from_effect(effect) for effect in self])
--- a/varcode/effect_collection.py
+++ b/varcode/effect_collection.py
@@ -268,8 +268,8 @@ class EffectCollection(Collection):
         """Build a dataframe from the effect collection"""
         def row_from_effect(effect):
             row = {}
-            row['chr'] = effect.variant.contig
-            row['position'] = effect.variant.start
+            row['contig'] = effect.variant.contig
+            row['start'] = effect.variant.start
             row['ref'] = effect.variant.ref
             row['alt'] = effect.variant.alt
             row['gene_id'] = effect.gene_id
--- a/test/test_effect_collection.py
+++ b/test/test_effect_collection.py
@@ -1,0 +1,26 @@
+# Copyright (c) 2016. Mount Sinai School of Medicine
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


iskandr:
Very slight nitpick: what do you think of `to_dataframe` instead of `as_dataframe`? It feels slightly more "pythonic" and seems to be the preferred method name for other projects: [BigQuery](http://googlecloudplatform.github.io/datalab/gcp.bigquery.html#gcp.bigquery.Query.to_dataframe), [django-pandas](https://pypi.python.org/pypi/django-pandas), [xarray](http://xarray.pydata.org/en/v0.7.1/generated/xarray.Dataset.to_dataframe.html?highlight=to_dataframe).

arahuja:
Ah yea, I like `to_dataframe` as well, but wanted to match the issue #150

tavinathanson:
Issue https://github.com/hammerlab/varcode/issues/128?

arahuja:
Yes, updated, I meant #128

varcode/effect_collection.py
```diff
@@ -262,3 +263,26 @@ def gene_counts(self):
...
+            row['alt'] = effect.variant.alt
+            row['gene_id'] = effect.gene_id
+            row['gene_name'] = effect.gene_name
+            row['transcript_id'] = effect.transcript_id
+            row['transcript_name'] = effect.transcript_name
+            row['variant'] = str(effect.variant)
+            row['is_indel'] = effect.variant.is_indel
+            row['is_transversion'] = effect.variant.is_transversion
+            row['is_transition'] = effect.variant.is_transition
+            row['is_transversion'] = effect.variant.is_transversion
```
iskandr: Twice?

varcode/effect_collection.py
```diff
@@ -262,3 +263,26 @@ def gene_counts(self):
...
+            row['chr'] = effect.variant.contig
+            row['position'] = effect.variant.start
+            row['ref'] = effect.variant.ref
+            row['alt'] = effect.variant.alt
+            row['gene_id'] = effect.gene_id
+            row['gene_name'] = effect.gene_name
+            row['transcript_id'] = effect.transcript_id
+            row['transcript_name'] = effect.transcript_name
+            row['variant'] = str(effect.variant)
+            row['is_indel'] = effect.variant.is_indel
```
iskandr: Worth adding 'is_snv' alongside 'is_indel?

varcode/effect_collection.py
arahuja: Woops, fixed

arahuja:
@iskandr seem to be running into the problem you have on pyensembl, where the py 2.7 tests on Travis on failing with ```The command "nosetests test --with-coverage --cover-package=varcode && ./lint.sh" exited with 137.```

iskandr:
@arahuja Should be fixed now but have to wait for Travis to run. The problem was that I was decoding byte strings into ASCII, which does the sane thing under Python 3 but under Python 3 switches from `str` to a fatter `unicode` representation which uses 2 (or 4!) times as much space per character. The big dictionaries of transcript sequences are about ~200-300mb each per release, so with 4x larger characters it's easy to imagine how we blow through Travis's 3gb limit.

arahuja:
@iskandr Looks to be passing now

arahuja:
@iskandr Good to merge?

iskandr:
@arahuja Definitely

varcode/effect_collection.py
```diff
@@ -262,3 +263,26 @@ def gene_counts(self):
         for effect in self:
             counter[effect.gene_name] += 1
         return counter
+
+    def to_dataframe(self):
+        """Build a dataframe from the effect collection"""
+        def row_from_effect(effect):
+            row = {}
+            row['chr'] = effect.variant.contig
+            row['position'] = effect.variant.start
```
timodonnell: might make sense to be clear here about the coordinate system, e.g. call it `inclusive_start`. Just a suggestion, feel free to ignore

varcode/effect_collection.py
```diff
@@ -262,3 +263,26 @@ def gene_counts(self):
         for effect in self:
             counter[effect.gene_name] += 1
         return counter
+
+    def to_dataframe(self):
+        """Build a dataframe from the effect collection"""
+        def row_from_effect(effect):
+            row = {}
+            row['chr'] = effect.variant.contig
```
timodonnell: I think `contig` might be better than `chr`. Also, I'd suggest including the reference genome in each row (e.g. a column called `genome` that gets set to `grch37` in every row)

timodonnell:
these are the column names i've been using. Not necessary to match them but just fwiw https://github.com/hammerlab/varlens/blob/master/varlens/variants_util.py#L26

varcode/effect_collection.py
arahuja: Yea, I think it's best to match the varcode field names

varcode/effect_collection.py
arahuja: Similarly, I'll make it match the varcode field names

PR #153, opened 2016-05-27
Reorganized coding effects to use KnownAminoAcidChange base class

iskandr:
Also added unit test for NoncodingTranscript effect

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/153)

### DIFF
--- a/test/test_effect_classes.py
+++ b/test/test_effect_classes.py
@@ -24,7 +24,7 @@ from varcode import (
     # transcript effects
     #
     IncompleteTranscript,
-    # NoncodingTranscript, TODO: write a noncoding transcript test
+    NoncodingTranscript,
     FivePrimeUTR,
     ThreePrimeUTR,
     Intronic,
@@ -62,6 +62,14 @@ def test_incomplete():
         transcript_id="ENST00000450046",
         effect_class=IncompleteTranscript)
 
+
+def test_noncoding_polymorphic_pseudogene():
+    # variant in MROH5-001, which is a polymorphic pseudogene
+    variant = Variant("8", 142458077, "C", "T", ensembl_grch37)
+    expect_effect(
+        variant,
+        transcript_id="ENST00000430863",
+        effect_class=NoncodingTranscript)
 
 def test_start_loss():
     # transcript EGFR-005 (ENST00000420316 in Ensembl 77)
--- a/test/test_effect_classes.py
+++ b/test/test_effect_classes.py
@@ -39,6 +39,7 @@ from varcode import (
     FrameShift,
     ExonLoss,
     ExonicSpliceSite,
+    FrameShiftTruncation,
     # TODO: SpliceDonor, SpliceReceptor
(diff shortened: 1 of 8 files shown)


test/test_effect_classes.py
```diff
@@ -75,7 +88,9 @@ def test_start_loss():
     expect_effect(
         variant,
         transcript_id="ENST00000420316",
-        effect_class=StartLoss)
+        effect_class=StartLoss,
+        modifies_coding_sequence=True,
+        modifies_protein_sequence=True)
```
arahuja: `StartLoss` modifies_protein_sequence? I guess it does, but tbh not what I would have expected when filtering on `modifies_protein_sequence` so good to know!

varcode/effects.py
```diff
@@ -382,13 +395,42 @@ def __init__(
         self.aa_mutation_end_offset = aa_mutation_end_offset
         self.aa_ref = aa_ref
 
-    modifies_protein_sequence = True
+    @property
+    def modifies_protein_sequence(self):
```
arahuja: In the tests you have 

```       
         effect_class=AlternateStartCodon,
         modifies_coding_sequence=True,
         modifies_protein_sequence=False)
```

varcode/effects.py
```diff
@@ -382,13 +395,42 @@ def __init__(
...
+        # TODO: scan downstream in the cDNA sequence to predict an alternate
+        # start codon using a Kozak sequence template
+        return None
+
+    @memoized_property
+    def short_description(self):
+        return "p.%s1? (start-loss)" % (self.transcript.protein_sequence[0],)
 
-class BaseSubstitution(NonsilentCodingMutation):
+class KnownAminoAcidChange(NonsilentCodingMutation):
```
arahuja: Does the `Known` part imply anything here? Why not just `AminoAcidChange`? This is everything that has `modifies_protein_sequence == True`?

iskandr: Effects such as `StartLoss` modify the protein sequence but we don't know what the amino acid change will be! So, this is really a base class for any effect with non-`None` mutant protein sequence.

varcode/effects.py
iskandr: Good catch, strange that the unit test doesn't fail.

iskandr: Actually, I don't see that on my local copy of the file.

test/test_effect_classes.py
iskandr: I'm imagining any variant you would expect to give you a different protein (even if its sequence is unknown) has `modifies_protein_sequence` as True. The only weird case I haven't figured out yet is `ExonLoss`.

varcode/effects.py
iskandr: Oh, the diff is confusingly mixing up the head of `AlternateStartCodon` and the tail of `NonsilentCodingMutation`

PR #155, opened 2016-05-28
Fix versioneer prefix format

armish:
The `cfg` syntax is tricky to get right :\ This should fix the versioneer for you.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/155)


iskandr:
Lgtm, merging

Issue #156, opened 2016-06-03
Memory leak due to excessive memoization


iskandr:
See: https://github.com/hammerlab/pyensembl/issues/150

iskandr:
I let Varcode run up about 1gb of memory doing effect prediction and then checked out the most allocated objects with [objgraph](https://www.huyng.com/posts/python-performance-analysis).

```
list                       853298
weakproxy                  360369
_Link                      360369
tuple                      313762
dict                       300230
OrderedDict                90092
Exon                       46690
Variant                    45124
function                   26384
Transcript                 14879
```

iskandr:
We get a slightly different summary from [pympler](https://pythonhosted.org/Pympler/muppy.html) but in either case I'm not sure how these objects can add up to >1gb. 

```
                             <class 'dict |      300318 |    126.16 MB
          <class 'collections.OrderedDict |       90092 |    112.04 MB
                              <class 'str |      748295 |     87.50 MB
                             <class 'list |      853389 |     82.99 MB
                        <class 'weakproxy |      360369 |     27.49 MB
                <class 'collections._Link |      360369 |     24.74 MB
                            <class 'tuple |      309318 |     24.59 MB
          <class 'varcode.variant.Variant |       45124 |      7.92 MB
                              <class 'int |      264699 |      7.09 MB
                             <class 'code |       24390 |      3.35 MB
                            <class 'float |      135719 |      3.11 MB
                             <class 'type |        2917 |      2.94 MB
              <class 'pyensembl.exon.Exon |       46690 |      2.49 MB
                              <class 'set |        2269 |    979.34 KB
  <class 'pyensembl.transcript.Transcript |       14879 |    813.70 KB
```

timodonnell:
Random thought... might consider a caching/memoization library with support for expiration like joblib http://pythonhosted.org/joblib/

hammer:
If we are going to do surgery here I'd rather see us push more logic down to the database rather than making our application-layer logic ever more arcane.

iskandr:
@hammer So far I'm focusing on lifting memoization where it's not needed, so hopefully making the code less rather than more arcane. Eager to hear suggestions about which parts might be easily pushed into the database.

iskandr:
@timodonnell It looks like the joblib memoization is specifically for writing results out to disk.

iskandr:
Old PR fixed this as much as I think is worth any effort: https://github.com/openvax/varcode/pull/157

PR #157, opened 2016-06-03
Don't memoize so much

iskandr:
Working on fixing https://github.com/hammerlab/varcode/issues/156, this doesn't clean up everything but it's a start.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/157)

### DIFF
--- a/varcode/vcf.py
+++ b/varcode/vcf.py
@@ -22,7 +22,7 @@ from six.moves import urllib
 
 import pandas
 from typechecks import require_string
-import vcf
+import vcf as pyvcf
 
 from .reference import infer_genome
 from .variant import Variant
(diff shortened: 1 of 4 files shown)


varcode/vcf.py
```diff
@@ -12,19 +12,17 @@
...
 import zlib
-import collections
+from collections import OrderedDict
 
 from six.moves import urllib
 
 import pandas
 from typechecks import require_string
-import vcf  # PyVCF
+import vcf
```
timodonnell: why drop this comment? it's kinda confusing otherwise i think

varcode/vcf.py
```diff
@@ -12,19 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# required so that 'import vcf' gets the global PyVCF package,
-# rather than our local vcf module
-from __future__ import absolute_import
```
timodonnell: why drop this comment?

iskandr: Since every module now has `absolute_import` so it's weird to explain why we're doing it in this case.

varcode/vcf.py
iskandr: Yeah, agreed, I'll add it back.

iskandr:
@timodonnell: OK to merge?

timodonnell:
yup LGTM

PR #159, opened 2016-07-11
Fix clone_with_new_elements for VariantCollection

tavinathanson:
Per https://github.com/hammerlab/cohorts/issues/66, we decided to use `clone_with_new_elements`; but I just realized that it wasn't actually cloning the `metadata` variable (just other "metadata").

@iskandr

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/159)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,9 +41,9 @@ script:
   - pyensembl install --release 77 --species human
   - pyensembl install --release 81 --species human
   # latest human Ensembl release
-  - pyensembl install --release 83 --species human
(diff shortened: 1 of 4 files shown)


tavinathanson:
The behavior of this method is somewhat weird, as the `metadata` dictionary is left alone, even if some keys become obsolete or some elements are no longer keyed by it. @iskandr suggested offline to make that behavior more clear, which I did.

tavinathanson:
cc @arahuja FYI

iskandr:
LGTM

PR #160, opened 2016-07-15
Fix load_vcf_fast for sample names containing a space character

### LINKED ISSUE #158: load_vcf_fast fails when sample names contain spaces
I was looking at a VCF where the sample name was "medullary thyroid" and got the following error from `load_vcf_fast`: 

```
CParserError: Too many columns specified: expected 11 and found 10
```

Using just `load_vcf` worked.

timodonnell:
This requires no longer accepting space-delimited VCF files, only tab-delimited files (which is what the VCF spec requires). Changed test/data/multiallelic.vcf to now be tab separated.

Closes #158

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/160)


timodonnell:
Travis was failing due to tests requiring pyensembl release 84 for human and mouse. Updated travis config to install those releases, will see if that fixes it

arahuja:
Just browsing through, don't know what the original issue was, but are there changes to `load_vcf_fast` here?

timodonnell:
@arahuja yeah the fix was to pass `strict_whitespace=True` when we make a pyvcf Reader. Not sure why the regular `load_vcf` was working but not `load_vcf_fast` actually. 

@iskandr can you confirm that your VCF with a space in the sample name can be loaded correctly now by both `load_vcf` and `load_vcf_fast`?

timodonnell:
Rebase onto latest master (for some reason this wasn't done before, sorry about that!)

iskandr:
Works for me! 

```
'sample_info': OrderedDict([('medullary thyroid',
```

PR #161, opened 2016-07-26
Commandline interface, simplified serialization, merging VariantCollections

iskandr:
* Added `union` and `intersection` to `VariantCollection`. 
* We previously allowed one unique path and one metadata dictionary for each `VariantCollection`. This complicates merging collections since they can now come from multiple sources. I replaced `path` with a more general notion of 'source' and the tied sources to metadata by having them both come from a parameter called `sources_to_metadata_dict`.  I tried to maintain backward compatibility by providing helper properties like `source` and `metadata` for the most common situation when we get variants from a single source.
* Simplified serialization to pickle and JSON by using the same `Serializable` base class as PyEnsembl. 
* Add `element_type` to `Collection`.
* Made `varcode-variants` script which loads multiple VCF/MAF/manually specified variants, takes their union and can write out a combined collection.
* Removed some cached properties which don't add much to performance but do eat up memory (general philosophy is now to only add caching around bottleneck methods, like those that talk to a database)

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/161)


varcode/vcf.py
```diff
@@ -230,15 +231,15 @@ def sample_info_parser(unparsed_sample_info_strings, format_string):
...
         info_parser=handle.vcf_reader._parse_info if include_info else None,
         only_passing=only_passing,
         max_variants=max_variants,
         sample_names=handle.vcf_reader.samples if include_info else None,
         sample_info_parser=sample_info_parser,
         variant_kwargs={
             'ensembl': genome,
-            'allow_extended_nucleotides': allow_extended_nucleotides},
-        variant_collection_kwargs={"path": path})
+            'allow_extended_nucleotides': allow_extended_nucleotides})
```
arahuja: Does `variant_collection_kwargs` no longer need to be passed?

varcode/vcf.py
```diff
@@ -296,6 +301,9 @@ def dataframes_to_variant_collection(
 
     variant_collection_kwargs : dict, optional
         Additional keyword parameters to pass to VariantCollection.__init__.
+
+    sources : set or list
```
arahuja: Is this is in the function def?

varcode/variant_collection.py
```diff
@@ -132,9 +138,9 @@ def clone_with_new_elements(self, new_elements):
         """
         return self.__class__(
             new_elements,
-            path=self.path,
+            source_to_metadata_dict=self.source_to_metadata_dict,
```
arahuja: Is this the same as the superclass version?

varcode/variant.py
```diff
@@ -247,83 +236,54 @@ def __eq__(self, other):
...
-        return instance
-
-    @classmethod
-    def from_json(cls, serialized):
-        """
-        Unserialize a Variant encoded as a string of JSON.
-        """
-        return Variant.from_dict(json.loads(serialized))
+        return dict(
+            contig=self.contig,
```
arahuja: Do you want the original contig too then?

varcode/collection.py
```diff
@@ -103,8 +191,9 @@ def clone_with_new_elements(self, new_elements):
         """
         return self.__class__(
             new_elements,
-            path=self.path,
-            distinct=self.distinct)
+            distinct=self.distinct,
```
arahuja: Need to pass along `element_type`?

varcode/collection.py
```diff
@@ -16,39 +16,130 @@
...
+        """
+        return self.source
+
+    @property
+    def filenames(self):
+        """
+        Assuming sources are paths to VCF or MAF files, trim their directory
+        path and return just the file names.
+        """
+        return [os.path.split(source)[1] for source in self.sources if source]
```
arahuja: Is this the same as `os.basename`?

varcode/cli/variants_script.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    if args_list is None:
+        args_list = sys.argv[1:]
+    arg_parser = make_variants_parser(
+        description="Annotate variants with overlapping gene names")
+    arg_parser.add_argument("--output-csv", help="Output path to CSV")
+    args = arg_parser.parse_args(args_list)
+    variants = variant_collection_from_args(args)
+    variants_dataframe = variants.to_dataframe()
+    print(variants_dataframe)
```
arahuja: Do you want to print this?

varcode/cli/variant_args.py
```diff
@@ -0,0 +1,139 @@
...
+
+    if args.genome:
+        genome = genome_for_reference_name(args.genome)
+    else:
+        # no genome specified, assume it can be inferred from the file(s)
+        # we're loading
+        genome = None
+
+    for vcf_path in args.vcf:
+        variant_collections.append(load_vcf_fast(vcf_path, genome=genome))
```
arahuja: Seems like this the default now? Worth dropping the other?

varcode/variant_collection.py
```diff
@@ -253,3 +202,97 @@ def exactly_equal(self, other):
...
+        """
+        return self._combine_variant_collections(
+            combine_fn=set.intersection,
+            variant_collections=(self,) + others,
+            kwargs=kwargs)
+
+    def to_dataframe(self):
+        """Build a DataFrame from this variant collection"""
+
+        def row_from_variant(variant):
```
arahuja: If you pull this out to a class level function, it can be re-used in `EffectCollection.to_dataframe()`

arahuja: Sort of ... I guess the gene handling is different

varcode/cli/variant_args.py
iskandr: Probably. @timodonnell Do you think the time has come?

varcode/cli/variants_script.py
iskandr: Yeah, that's intentional. I like seeing the result even if nothing gets written to a file.

varcode/vcf.py
iskandr: Not any more, `dataframes_to_variant_collection` populates metadata & sources on its own.

varcode/vcf.py
iskandr: Ah forgot about this. Curious what you think about the presence of `sources_to_metadata_dict` on `EffectCollection`. It seems confusing since the metadata dictionaries are still keyed by `Variant`. Maybe the base `Collection` class should only know about a set of sources and only `VariantCollection` has a full `sources_to_metadata_dict`?

varcode/collection.py
iskandr: Good catch.

varcode/variant.py
iskandr: Also a good catch. Excited to eventually demote "normalization" to properties and have start/ref/alt be the same values as in a VCF.

varcode/variant_collection.py
iskandr: What do you mean?

iskandr: Yeah, I think the difference seemed reasonable when I was considering having `Collection` only take `sources` but not `source_to_metadata_dict`.

varcode/collection.py
iskandr: Yep :-)

varcode/vcf.py
arahuja: I'd say it's useful? A common operation for us in `cohorts` is just saving/loading the `EffectCollection` and not always having the `VariantCollection` around. I think they should be able to stand on their own, but obviously if you take it out we can reference the variant collection

varcode/cli/variant_args.py
arahuja: Looks like it is deprecated now, so worth swapping this one?

varcode/vcf.py
iskandr: I think I'm going to leave it out, since the metadata dictionary has gotten somewhat complicated and I'd like to localize that complexity to VariantCollection

Issue #181, opened 2016-09-19
Reference incorrectly inferred when "b36" in reference file path


jburos:
Seems like some reference genomes are incorrectly inferred when there is a partial match in the file path.

Example:
```
In [39]: varcode.reference.infer_reference_name('##reference=file:///var/lib/cwl/job367935311_index_001zdr/GRCh38.d1.vd1.fa')
Out[39]: 'NCBI36'
```

This is a not-uncommon pattern in files downloaded from the GDC. 

The relevant code is https://github.com/hammerlab/varcode/blob/master/varcode/reference.py#L51-L55

Two suggestions:

1. use only the filename, not the path to infer reference (per @arahuja)
2. modify sorting of the `reference_alias_dict` so that GRCh38, GRCh37, etc are checked before NCBI36.

PR #182, opened 2016-09-19
Fix inferred-reference-bug

jburos:
See issue #181 -- reference is sometimes incorrectly attributed to NCBI36 when (for example) "b36
 is in the file path.

This PR attempts to address that by hard-coding the sort order of references to check, so that search looks for GRCh38 before GRCh37 (for example) & looks for human references before mouse references.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/182)

### DIFF
--- a/varcode/reference.py
+++ b/varcode/reference.py
@@ -68,24 +68,22 @@ def infer_reference_name(reference_name_or_path):
                 matches['file_name'].append(assembly_name)
             elif candidate.lower() in reference_name_or_path.lower():
                 matches['full_path'].append(assembly_name)
+
     # given set of existing matches, choose one to return
     # (first select based on file_name, then full path. If multiples, use most recent)
     if len(matches['file_name']) == 1:
         match = matches['file_name'][0]
     elif len(matches['file_name']) > 1:
         # separate logic for >1 vs 1 to give informative warning
-        most_recent = _most_recent_assembly(matches['file_name'])
-        warn('More than one reference matches path in header ({path});' +
-            ' the most recent one ({selected}) was used.'.format(
-                path=reference_file_name, selected=most_recent))
-        match = most_recent
+        match = _most_recent_assembly(matches['file_name'])
+        warn('More than one reference ({}) matches path in header ({});' \
+            .format(','.join(matches['file_name']), reference_file_name) +
+            ' the most recent one ({}) was used.'.format(match))
     elif len(matches['full_path']) >= 1:
         # combine full-path logic since warning is the same
-        most_recent = _most_recent_assembly(matches['full_path'])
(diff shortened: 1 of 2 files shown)


varcode/reference.py
```diff
@@ -25,19 +24,22 @@
...
+    "GRCm37": {'aliases': ["mm9"], 'sort_order': 1.37},
+    "GRCm38": {'aliases': [
         "mm10",
         "GCF_000001635.24",  # GRCm38.p4
         "GCF_000001635.23",  # GRCm38.p3
         "GCF_000001635.22",  # GRCm38.p2
         "GCF_000001635.21",  # GRCm38.p1
         "GCF_000001635.20",  # GRCm38
-    ],
+    ], 'sort_order': 1.38}
```
iskandr: Can you explain more about what a sort order of `1.38` means?

iskandr:
The current code is pretty fragile but I'm worried that fixing it via an enforced sort order leaves us exposed to the same kinds of bugs. What if a GRCh37 VCF had a reference path with "job38" in it? 

Maybe one alternative is to (1) try inferring from filename before using the directory and (2) only try looking for reference names at the start of strings between separators (e.g. "grch38p2.fasta") but don't allow open ended string containment (i.e. "b37" shouldn't match "job37").  

This is a bunch of hacky heuristics but hopefully will work a larger fraction of the time.

jburos:
This makes a lot of sense, thanks for reviewing. I wonder if it is
reasonable to infer the reference from the directory at all, if the file
doesn't match? Also, perhaps we should give a warning if either we see more
than one possible match, or no match on file name.

iskandr:
LGTM except that such tricky-looking logic deserves a few unit tests.

jburos:
Yeah - you're absolutely right. I will put together a few test input
strings, with the idea that it can be extended as additional problem
references are found.

Is it worth maybe splitting off any of these elements into separate
functions? Ie, the part that picks the "most recent", or the matching on
dict + aliases?

You'll note I didn't implement the `re` to match the first part of the
filename; figured I should get this logic down before adding that element.

Thx for the feedback!

iskandr:
It sure seems like the repeated recency code could benefit from being pulled out into a helper function (and then tested in isolation). 

Also, I think that you might get away with not needing the recency logic if you instead implement some sort of delimiter splitting over parts of the filename. But I'm happy with any solution that solves your `job36` problem and is easy to extend/modify later.

iskandr:
@jburos Just checking in, how is this branch doing?

jburos:
@iskandr I added a few test cases & separated the recency bit into a function as recommended. 

Will wait to see how tests finish. Thanks & LMK your thoughts --

PR #185, opened 2016-09-27
Upgrade serializable dependency with tests

arahuja:
- upgraded `serializable` dependency
- added a few tests for serializing merged variant collections that were failing before

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/185)


iskandr:
LGTM 

Nothing that we should probably also try pickling individual instances of all the distinct effect classes.

Issue #202, opened 2016-12-01
Variants before and over the stop codon lose information.


joaoe:
Howdy.

Testcase, over transcript ENSMUST00000086738, GRCm38
```
chr1 100484693 . TTCATCTGA CCC
```
The end of that transcript looks like
```
TTC ATC TGA ACT
 F   I   *   T
```
and this mutation will cause the location plus downstream to become
```
PTIVWSSGPLF(...)
```
The annotation that varcode gives is
```
StopLoss
* aa_mutation_start_offset = 1292
* aa_ref="*"
* aa_alt="PTIVWSS(...)"
```
It should actually be
```
StopLoss
* aa_mutation_start_offset = 1290
* aa_ref="FI*"
* aa_alt="PTIVWSS(...)"
```
The main issue is that even though `aa_ref` could be disregarded, `aa_mutation_start_offset` and `aa_alt` are desynchronized, since `aa_alt` actually starts two AAs upstream.

My suggestion is just to add `start, aa_ref, aa_alt`to the constructor of `StopLoss`like all other variants. `PrematureStop`is also about fiddling with stop codons and it contains those properties.

iskandr:
Good catch! Thanks for continuing to come up with good test cases.

iskandr:
Fixed by https://github.com/openvax/varcode/pull/203/files

PR #203, opened 2016-12-03
Adding aa_ref argument to StopLoss for variants which delete codons before stop

iskandr:
Issue: https://github.com/hammerlab/varcode/issues/202

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/203)

### DIFF
--- a/varcode/effects/effect_classes.py
+++ b/varcode/effects/effect_classes.py
@@ -617,8 +617,8 @@ class StopLoss(KnownAminoAcidChange):
             self,
             variant,
             transcript,
-            extended_protein_sequence,
-            aa_ref="*"):
+            aa_ref,
+            aa_alt):
         # StopLoss assumes that we deleted some codons ending with a
         # stop codon
         if not aa_ref.endswith("*"):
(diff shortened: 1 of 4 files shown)


varcode/effects/effect_classes.py
```diff
@@ -617,14 +617,24 @@ def __init__(
             self,
             variant,
             transcript,
-            extended_protein_sequence):
-        aa_mutation_start_offset = len(transcript.protein_sequence)
+            extended_protein_sequence,
```
joaoe: To keep it in syc with the other classes, perhaps having aa_ref/aa_alt in that order would make more sense.

iskandr: Good idea, what do you think of the current changes? I'm a little nervous about having '*' in the aa_ref sequence since the stop codon isn't made explicit anywhere else (e.g. it's not in truncating effects).

joaoe: First time I saw `"*"` in `aa_ref` I was like "what just crashed my script?" (handling a big vcf file, generating mutated protein sequences to a file).

I don't think it's necessary to have the stop codon there, since it's implied that `StopLoss` already affects it.

Being pedantic if you have a variant that affects both sides of the stop codon (e.g. `chr1 100484693 TTCATCTGAACT    CCCTAACCCCCC`), then `aa_ref` would be `FI*T`.

Issue #210, opened 2017-01-18
Allow contig to be empty to allow parsing of MAF with faulty mutation(s)


tuomastik:
I'm proposing `varcode.variant.Variant()` to allow (with warning message) an empty value in `contig` attribute instead of raising an exception. This way user will be able to use `load_maf` parser with faulty MAF where a mutation has an empty value in "Chromosome" field.

Current functionality:
```
>>> variant_collection = varcode.load_maf('missing_chr.maf')
TypeError: Chromosome cannot be 'nan' : <class 'float'>
```

Proposed functionality:
```
>>> variant_collection = varcode.load_maf('missing_chr.maf')
Warning: Something went wrong when normalizing chromosome: Chromosome cannot be 'nan' : <class 'float'>
```

iskandr:
Huh, what's the use of a variant which lacks a chromosome? Wouldn't you rather catch the exception in `load_maf` and just skip that variant?

tuomastik:
I'm using the MAF parser in a project where MAF files are provided by user. Thanks for the great work by the way. 👍 However, I recently ran into a user supplied MAF which contained thousands of perfectly formatted variants except for one that had a missing value in "Chromosome" field. Because of that one erroneous variant the whole parsing failed which I would now like to avoid by allowing (not skipping) variants lacking a chromosome and manually checking for missing chromosomes after the parsing is done.

joaoe:
The `Variant.effects` method already accepts a `raise_on_error` parameter. You could use the same principle for load_vcf and load_maf. Something like:
```
# raise - raise exception when parsing broken variant
# keep broken variant object
# skip broken variant and do not create Variant object.
load_maf(..., on_error = ('raise', 'keep', 'skip'))
```

iskandr:
@tuomastik Would you want to skip the variants with missing chromosomes or use external knowledge to fill them in? I'd like to fix your problem without allowing the construction of malformed `Variant` objects. 

Also, sorry for the slow response~ (cycling between different projects I guess leads to long lag times)

tuomastik:
@iskandr I would not like to skip variants with missing chromosomes. I would be happy with malformed `Variant` objects because I'm passing the Varcode parsed variants through a custom validator. I understand that allowing malformed `Variant` objects is not something that you would like to allow, but actually @joaoe's suggestion sounds like a good middle ground solution to me, letting users to optionally allow malformed `Variant` objects.

PR #221, opened 2017-04-01
added raise_on_error option to load_maf and load_maf_dataframe

iskandr:
Fixes https://github.com/hammerlab/varcode/issues/210 for @tuomastik

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/221)

### DIFF
--- a/varcode/maf.py
+++ b/varcode/maf.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2016. Mount Sinai School of Medicine
+# Copyright (c) 2016-2017. Mount Sinai School of Medicine
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -13,10 +13,11 @@
 # limitations under the License.
 
 from __future__ import print_function, division, absolute_import
+import logging
 
 import pandas
-
 from typechecks import require_string
+from numpy import isnan
 
 from .reference import infer_genome
 from .variant import Variant, variant_ascending_position_sort_key
@@ -47,9 +48,20 @@ MAF_COLUMN_NAMES = [
 ]
 
 
-def load_maf_dataframe(path, nrows=None, verbose=False):
+def load_maf_dataframe(path, nrows=None, raise_on_error=True):
     """
     Load the guaranteed columns of a TCGA MAF file into a DataFrame
+
+    Parameters
+    ----------
+    path : str
+        Path to MAF file
+
+    nrows : int
+        Optional limit to number of rows loaded
+
+    raise_on_error : bool
+        Raise an exception upon encountering an error or log an error
     """
     require_string(path, "Path to MAF")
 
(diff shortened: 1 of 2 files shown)


varcode/maf.py
```diff
@@ -66,9 +78,13 @@ def load_maf_dataframe(path, nrows=None, verbose=False):
...
 
     if len(df.columns) < n_basic_columns:
-        raise ValueError(
+        error_message = (
             "Too few columns in MAF file %s, expected %d but got  %d : %s" % (
                 path, n_basic_columns, len(df.columns), df.columns))
+        if raise_on_error:
+            raise ValueError(error_message)
+        else:
+            logging.warn(error_message)
```
tavinathanson: What happens here, with the wrong number of columns, when we don't error?

varcode/maf.py
```diff
@@ -82,38 +98,60 @@ def load_maf_dataframe(path, nrows=None, verbose=False):
...
         raise ValueError("Empty MAF file %s" % path)
 
     ensembl_objects = {}
     variants = []
     metadata = {}
     for _, x in maf_df.iterrows():
         contig = x.Chromosome
+        if not contig or isnan(contig):
+            error_message = "Invalid contig name: %s" % (contig,)
+            if raise_on_error:
```
tavinathanson: Looks like we're not keeping any malformed variants, which I agree with. Maybe these warning messages should clarify that we're skipping the variant; otherwise could be assumed that they're just malformed or something else?

iskandr:
@tavinathanson The trouble with adding it to `load_vcf` is that the code there is significantly more complicated with many more opportunities for errors. I think I'd rather hold off on that until there's a user request.

varcode/maf.py
iskandr: Good idea

varcode/maf.py
```diff
@@ -82,38 +98,60 @@ def load_maf_dataframe(path, nrows=None, verbose=False):
...
+        Raise an exception upon encountering an error or just log a warning.
     """
     # pylint: disable=no-member
     # pylint gets confused by read_csv inside load_maf_dataframe
-    maf_df = load_maf_dataframe(path)
+    maf_df = load_maf_dataframe(path, raise_on_error=raise_on_error)
 
-    if len(maf_df) == 0:
+    if len(maf_df) == 0 and raise_on_error:
         raise ValueError("Empty MAF file %s" % path)
```
iskandr: I think that if the user turned off errors then they're OK with getting back empty variant collections

varcode/maf.py
```diff
@@ -82,38 +98,60 @@ def load_maf_dataframe(path, nrows=None, verbose=False):
                 df[expected] = df[actual]
                 del df[actual]
             else:
-                raise ValueError("Expected column %s but got %s" % (
-                    expected, actual))
+                error_message = (
+                    "Expected column %s but got %s" % (expected, actual))
+                if raise_on_error:
```
joaoe: This pattern
```
if raise:
  raise ...
else:
  log();
```
is now quite common.
The code could be made cleaner with a helper function in the local scope
```
def raise_or_log(message):
  if (raise_on_error):
    raise ValueError(message)
  else
  logging.warn(message)
```

PR #222, opened 2017-04-04
Adding 'distinct' as a parameter to load_vcf.

julia326:
Default behavior remains the same (distinct=True).

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/222)


timodonnell:
LGTM

PR #223, opened 2017-04-28
Fixes to load_maf

tavinathanson:
In this PR:

* Pass `encoding` through; I had some MAFs that needed `encoding="latin1"` when in Python 3, for example. Add a test to ensure that this does not break Python 2 (was seeing `Bio.Seq` errors if I did not convert `ref` and `alt` to `str` in Python 2).
* Use `isnull` rather than `isnan`; the latter errors when chromosomes are strings vs. numerical (`TypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''`). Added a test for this.

Curious if I can handle any of the above more intelligently?

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/223)


tavinathanson:
@iskandr can you think of any reason I should have left `if not contig` vs. just using `isnull`?

PR #226, opened 2017-08-01
WIP: Basic VCF output functionality, given a single variant source.

jlkravitz:
Currently only works when variants come from a single source (i.e., this will break if you run `varcode-variants --vcf x.vcf --vcf y.vcf`).

Many TODOs thrown throughout the code, some of which I want to finish before considering this done for now – especially allowing for multiple variant sources.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/varcode/226)


timodonnell:
Will be cool to have VCF output functionality, thanks for working on this @jlkravitz (didn't realize this was one of your projects). Haven't gone through the code but two high level thoughts:
 * I would suggest moving the code to its own module, e.g. `vcf_output.py`
 * Probably want to start thinking about tests. I would suggest putting together a library of VCFs and verifying that they can be round-tripped (e.g. test equality of the variant collections you get by reading in the original VCF vs. what you get from reading it in, writing it out, and reading that in again). A starting point are the VCFs in the test dir https://github.com/hammerlab/varcode/tree/master/test/data

jlkravitz:
@timodonnell Thanks for the feedback! I'm generally just working on some smaller software projects as I learn more and work on data analysis stuff (this is one of those projects).

* Own module makes a lot of sense -- I have lots of nested helpers here that would be more readable in their own module.
* Re: testing. I've been testing on those, but would you recommend writing actual tests in a .py file? Also, would it be reasonable to make a folder `test/data/vcf-output` for tests specific to this module? I can imagine writing lots of little test files and combinations, but don't really want to clutter the primary test files folder.

timodonnell:
Yeah, definitely best to put actual tests in a `.py` file, similar to how the other functionality is tested in varcode. That way we get automated unit tests by just running `nosetests .` and in travis.

Also fine to make a folder for testing data as you mentioned. For the round-trip tests, I wouldn't check into git the files output by the test though, only the input files. In general I wouldn't worry about cluttering the primary test folder too much -- that's an easy problem to fix later if it ever becomes one.

jlkravitz:
@timodonnell @julia326 

I _think_ this is ready for basic usage. There are lots of additions to be made, most (or all?) of which I've written up in detail as `TODO`s in the code itself. (Some of these `TODO`s are questions, rather than specifications for how to improve what's here -- I'm not familiar enough with VCF usage to know which corner cases are important to handle and which are not.)

I also haven't written a "merge" function for merging metadata dictionaries, but figure users can just write their own for now. (I do have a naive implementation in the `test_vcf_output.py`, which just overwrites any duplicates.)

Let me know if you have any comments!

(Also, what's the general guideline re: commits? Should I squash all these commits -- as I did for the first one -- and have one clean commit for this PR? Or is it fine to just keep the mess?)

PR #235, opened 2018-10-09
fixed top_priority_effect for ExonicSpliceSite

iskandr:
Currently, if an `ExonicSpliceSite` effect has an `alternate_effect` which is more interesting (e.g. `StopGain`), it doesn't come up as the top priority. This changes the `top_priority_effect` logic to return the `alternate_effect` if it's higher priority than a potential exonic splice site change. 

Other changes:
* moved splicing mutations to be right after `Silent` in the effect ordering. 
* fixed bug I noticed in VCF writing where a random hg19 FASTA path was always used as the reference 

Fixes: https://github.com/openvax/varcode/issues/233

### DIFF
--- a/varcode/effects/effect_ordering.py
+++ b/varcode/effects/effect_ordering.py
@@ -53,9 +53,16 @@ transcript_effect_priority_list = [
     # so give 5' UTR mutations higher prioriry
     FivePrimeUTR,
     Silent,
-    # changing the stop codon might have more significance than
-    # silent variants internal to the coding sequence
-    # REMOVED: AlternateStopCodon,
+    # intronic variants near the splice boundaries but which aren't
+    # the two nucleotides closest to the exon
+    IntronicSpliceSite,
+    # exonic variants near a splice boundary
+    ExonicSpliceSite,
+    # modification or deletion of stop codon
+    SpliceDonor,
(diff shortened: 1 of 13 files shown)


varcode/effects/effect_ordering.py
```diff
@@ -125,13 +124,31 @@ def effect_sort_key(effect):
     effect_class_priority = effect_priority(effect)
     return (effect_class_priority, cds_length, transcript_length)
 
+
+def select_between_exonic_splice_site_and_alternate_effect(effect):
```
julia326: can you add some comments here? this functionality is kinda non-obvious

varcode/vcf_output.py
```diff
@@ -248,9 +249,11 @@ def get_sample_names():
         headers += ['FORMAT'] + sample_names
 
     unique_variants_list = merge_duplicate_variants()
+    # can't have more than one reference in VCF so pick the first one and pray
```
julia326: instead of praying, throw an error? when would we ever want to support combining VCFs created against different references?

iskandr: Changed it to check for a unique reference

varcode/effects/effect_ordering.py
iskandr: Added a docstring

PR #237, opened 2019-04-26
Smarter selection of top priority effect

iskandr:
So far we've used very simple logic to choose the top effect in an EffectCollection, which is taking the `max` with a sort key with three elements: 

1) priority of effect type
2) CDS length
3) full transcript length

This sort criterion seems to break down in some edge cases, such as https://github.com/openvax/varcode/issues/236 where a weird isoform of UBE2I has a longer CDS despite being shorter and losing several exons. This is not the canonical or expected isoform of UBE2I for a variant annotation -- its protein sequence is significantly different. 

I rewrote the `effect_ordering` to use a more fine grained process of filtering and sorting with additional heuristics to make sure a sane transcript gets chosen. 

When effects originate from different genes they are sorted according to the following criteria:

        1) Integer priority of the effect type.
        2) Does the associated gene have a "protein_coding" biotype?
            False if no gene is associated with effect.
        3) Does the associated transcript have a "protein_coding" biotype?
            False if no transcript is associated with effect.
        4) Is the associated transcript complete?
            False if no transcript is associated with effect.
        5) CDS length
             This value will be 0 if the effect has no associated transcript
             or if the transcript is noncoding or incomplete
        6) Total length of the transcript
             This value will be 0 intra/intergenic variants effects without
             an associated transcript.
        7) Number of exons
             This value will be 0 intra/intergenic variants effects without
             an associated transcript.
        8) If everything is the same up this point then let's use the very
           sloppy heuristic of preferring transcripts like "TP53-201" over
           "TP53-206", so anything ending with "01" is considered better.
        9) Lastly, if we end up with two transcripts like "TP53-202" and
           "TP53-203", prefer the one with the lowest number in the name.

For effects which are from the same gene, try to find a canonical transcript with both the maximum length CDS and maximum length full transcript sequence.

If there are conflicting transcripts (some with max CDS length, some with max full length) then consider them all for a final tie-breaking sort. 

Tie-breaking criteria for effects arising from the same gene: 
   1) number of exons
   2) does transcript name end with "01"
   3) prefer transcript names that end with lower numerical values

Other changes:

* Updated Ensembl release used for tests to 95
* Added `varcode` command for simplified interactive debugging
* Renamed `varcode-variants` to `varcode-genes`


.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,15 @@
+<component name="InspectionProjectProfileManager">
```
timodonnell: intentional to have this file?

varcode/cli/effects_script.py
```diff
@@ -0,0 +1,54 @@
...
+            --vcf strelka.vcf \
+            --maf tcga_brca.maf \
+            --variant chr1 498584 C G \
+            --json-variants more_variants.json
+    """
+    print_version_info()
+    if args_list is None:
+        args_list = sys.argv[1:]
+    arg_parser = make_variants_parser(
+        description="Annotate variants with overlapping gene names and predicted coding effects")
```
timodonnell: May want to include `main.__doc__` in the usage info so the users of the script get the the docstring you wrote here

.idea/inspectionProfiles/Project_Default.xml
iskandr: Opinions on having PyCharm project files in the repo, yay or nay?

varcode/cli/effects_script.py
iskandr: Can functions reference their own docstring?

PR #242, opened 2019-09-26
Correct handling of UCSC vs. Ensembl reference genomes

iskandr:
When we wrote PyEnsembl and Varcode we had an annoying problem: UCSC rolls their own slightly different reference genomes and names all the chromosomes differently (e.g. "chr1" instead of "1"). The hacky band-aid to get around this problem, so that we could freely mix hg19 and GRCh37 aligned variants, was to trim the "chr" prefix and in the special case of the mitochondrial contig rename "chrM" to "MT". 

This worked most but not all of the time and moved a very ugly detail in PyEnsembl, whose logic is otherwise mostly agnostic to stuff like this. The serialization of VCFs originating from UCSC references was also wrong since it would turn them into Ensembl coordinates and reference names. 

The current version of PyEnsembl has been updated to do universal normalization (e.g. stripping off whitespace and converting integers to strings) but without any UCSC specific logic. 

This PR makes the UCSC specific conversion more explicit by adding several fields to Variant:
* `original_genome`: what the genome was before possible conversion from from UCSC to Ensembl
* `original_reference_name`: name of reference genome before possible UCSC to Ensembl conversion
* `convert_ucsc_contig_names`: option which can be passed by the user but more often inferred from whether `original_reference_name` is in a list of known UCSC references. 

In addition to this I'm also adding some sanity checking preceding annotation (calling `variant.effects()`) which makes sure that whatever normalized contig name we end up with is also in the reference genome. The unfortunate previous behavior was just to annotate variants which had wrong contig names as "intergenic". 

Alt and unplaced contig names from UCSC still don't get correctly converted to GRCh37 names but this was true before: now at least they're left alone instead being converted to something non-existent. 

**Breaking change**: you can't do `Variant("chr1", pos, ref, alt, "GRCh37")` anymore and still have it work. Now you have to explicitly say `Variant("chr1", pos, ref, alt, "hg19")` if you're using UCSC contig names.


timodonnell:
LGTM after a quick look. Only real comment - the terms "genome" and "ensembl" seem to be used sort of interchangeably but maybe not quite identically (as e.g. parameter names) and it would be helpful to document somewhere (maybe in a class or module docstring) what exactly is meant by these different terms throughout the codebase. I'm assuming "ensembl" is a particular ensembl release, which corresponds to one genome, but I didn't see that spelled out anywhere.

Also not worth delaying this PR here but just to clarify - would an alternate possible interface be to add a `VariantCollection.liftover(self, genome) -> VariantCollection` method that could be used e.g. to convert a `hg19` VariantCollection to a `grch37` collection, which then has an associated ensemble release? That strikes me as a potentially cleaner implementation (putting liftover code all in one place) and more explicit interface in the long run, and even if initially it doesn't support anything non-trivial (e.g. only supports hg19 -> grch37 and drops MT variants with a warning). Then perhaps in `load_vcf` or elsewhere there could be an option to do automatic liftover to whatever the most similar genome is that has an associated ensembl release.  Just a thought though, and not worth delaying the current PR over.

iskandr:
Thanks @timodonnell -- I'm leaving `ensembl` as a deprecated alias for `genome` but it means the same thing. 

As far as the `liftover` idea: that's exactly what I wanted to do! But, it does break compatibility since downstream code uses implicit equivalence between `hg19` and `GRCh37` both in the tests and e.g. Vaxrank. 

So, I'm thinking of this PR as a halfway point where the hg19/GRCh37 logic is made a little more explicit and later I'd like to do a coordinated set of changes where Varcode gets a `liftover` method and all the code we have that relies on Varcode will be updated at once.

PR #253, opened 2024-07-03
Defaulting convert_ucsc_contig_names to True for consistency with past behavior

julia326:


iskandr:
👍 once tests pass

iskandr (approved):

iskandr:
@julia326 Noting that coveralls probably needs to be configured for parallel jobs but still merging this

Issue #309, opened 2026-04-15
[#271 stage 3d] SequenceDiffEffectAnnotator: classify effects from protein diff


iskandr:
Algorithmic stage of #271, split out from the #304 staging table. Plumbing (3a–3c) landed in 2.6.0; this issue tracks the actual sequence-diff classifier.

## Goal

Add a second :class:\`EffectAnnotator\` — \`SequenceDiffEffectAnnotator\` — that produces byte-for-byte identical output to \`LegacyEffectAnnotator\` on a curated validation corpus, but does so by:

1. Materializing the mutant transcript with \`apply_variant_to_transcript\`.
2. Comparing the translated mutant protein to the reference protein.
3. Classifying the effect from the minimal edit rather than reasoning about CDS offsets.

Legacy stays the default until stage 3e flips it. Until then, sequence_diff is an opt-in registered under \`\"sequence_diff\"\` and exercisable via \`variant.effects(annotator=\"sequence_diff\")\` or \`use_annotator(\"sequence_diff\")\`.

### 1. \`SequenceDiffEffectAnnotator\` class

Lives in \`varcode/annotators/sequence_diff.py\`. Shape:

\`\`\`python
class SequenceDiffEffectAnnotator:
    name = \"sequence_diff\"
    version = _varcode_version
    supports = frozenset({\"snv\", \"indel\", \"mnv\"})

    def annotate_on_transcript(self, variant, transcript):
        # Gate: non-coding / incomplete transcripts -> classical wrappers
        if not transcript.is_protein_coding:
            return NoncodingTranscript(variant, transcript)
        if not transcript.complete:
            return IncompleteTranscript(variant, transcript)

        # Shared fast path (stage 3c) — same as legacy dispatches through.
        fast = _try_fast_path_if_applicable(variant, transcript)
        if fast is not None:
            return fast

        # Splice effects stay the legacy classifier's job; sequence_diff
        # only owns protein-level diff classification.
        if _variant_is_splice_adjacent(variant, transcript):
            return LegacyEffectAnnotator().annotate_on_transcript(
                variant, transcript)

        # Slow path: materialize and diff.
        mt = apply_variant_to_transcript(variant, transcript)
        if mt is None:
            # UTR / ref-mismatch / out-of-single-exon — delegate.
            return LegacyEffectAnnotator().annotate_on_transcript(
                variant, transcript)
        return classify_from_protein_diff(mt, variant, transcript)
\`\`\`

[...]

Issue #305, opened 2026-04-15
Rewrite splice_outcomes.py on MutantTranscript (closes #262 prototype loop)


iskandr:
Spun out of the #271 staging discussion in #304. The \`splice_outcomes\` prototype (shipped in PR #292 as part of #262) is a substantial enough rewrite to deserve its own tracking issue rather than being a sub-stage of #271.

## Motivation

\`varcode/splice_outcomes.py\` today (~800 LOC) carries four separate pieces of ad-hoc sequence math that each exist because \`MutantTranscript\` wasn't available when the prototype shipped:

- \`_build_in_frame_exon_skip_effect\` — manually reconstructs the boundary codon when an exon's cDNA start sits at phase 1 or 2, then trims shared flanks to classify as Deletion / Substitution / ComplexSubstitution. This was the entire payload of #298.
- \`_build_out_of_frame_exon_skip_effect\` — builds a post-skip cDNA, translates to first stop, wraps the result in a private \`_ExonSkipFrameshiftEffect\` shim that can't be serialized through the normal path.
- \`_build_start_loss_effect\` — special-cases the skipped-exon-contains-start-codon scenario.
- \`_translate_to_first_stop\` — re-implements translate() with its own truncation logic because the main translator wasn't codon-table-aware when this was written (#294 fixed the main translator; this duplicate path persists).

All four are locally correct but they duplicate logic that \`apply_variant_to_transcript\` + \`MutantTranscript\` already handle. Rewriting \`splice_outcomes\` on top of the abstraction:

- removes the four ad-hoc builders
- gets boundary-codon reconstruction for free (it falls out of protein diff, same way #298 would have if it had landed post-#271)
- closes the #262 multi-effect-splice prototype loop — the module docstring currently calls itself \"an explicit prototype\" waiting on \`MutantTranscript\`
- unlocks intron retention / cryptic splice protein computation once genomic FASTA support lands (#296) — those paths will just build a \`MutantTranscript\` with the retained-intron / cryptic-boundary edits

## Scope

Each \`SpliceOutcome\` becomes a \`MutantTranscript\`:

| Outcome | Edits applied to reference transcript |
|---|---|
| \`NORMAL_SPLICING\` | The underlying coding variant (if any) |
| \`EXON_SKIPPING\` | Single deletion spanning the skipped exon's cDNA range |
| \`INTRON_RETENTION\` | Insertion of the intron's genomic sequence at the donor boundary (requires #296) |
[...]

Issue #304, opened 2026-04-15
[#271 stage 3 planning] SequenceDiffAnnotator design, naming, and reordered staging


iskandr:
Planning issue for the next stages of #271, drafted after stages 1 (data model + Protocol + LegacyEffectAnnotator) and 2 (apply_variant_to_transcript) landed. Locks in naming, outlines the classifier algorithm, and proposes a staging reorder.

## Naming

**Proposed**: rename the Protocol and classes to drop the redundant \"Effect\" prefix that I added in stage 1.

| Current (2.5.0) | Proposed |
|---|---|
| `varcode.annotators.EffectAnnotator` (Protocol) | `varcode.annotators.Annotator` |
| `varcode.annotators.LegacyEffectAnnotator` | `varcode.annotators.LegacyAnnotator` |
| *(not yet)* | `varcode.annotators.SequenceDiffAnnotator` |

Rationale:
- The module is already `varcode.annotators`; `EffectAnnotator` re-states the concept.
- The issue spec uses path-style identifiers `varcode.annotators.legacy` / `varcode.annotators.sequence_diff` — the class names should read cleanly next to them.
- Cheaper to rename now (one public class + a few tests) than after a second class lands and doubles the migration.

**Back-compat**: keep `EffectAnnotator` and `LegacyEffectAnnotator` as aliases for one minor version, deprecate via docstring, remove in the version after.

**Alternative**: leave names as-is. Slightly verbose but zero migration cost. Open to either — flagging so the choice is deliberate.

## Will legacy stay around?

Yes. The `LegacyAnnotator` is the existing `predict_variant_effect_on_transcript` path already wrapped as an object (stage 1). It stays registered and available after `sequence_diff` becomes the default, because:

1. Third-party pipelines pin varcode versions expecting byte-for-byte stable output.
2. A/B comparison between annotators requires both to coexist.
3. Removing legacy is a separate decision with its own deprecation cycle.

So the question isn't \"replace legacy with sequence_diff\" — it's \"add a second annotator behind the same interface, let users opt into either.\"

## SequenceDiffAnnotator algorithm

Skeleton:

```python
class SequenceDiffAnnotator:
    name = \"sequence_diff\"
    supports = frozenset({\"snv\", \"indel\", \"mnv\"})

    def annotate_on_transcript(self, variant, transcript):
        # 1. Gate: non-coding / incomplete transcripts go straight to legacy.
        if not transcript.is_protein_coding:
[...]

Issue #299, opened 2026-04-14
Unified MultiOutcomeEffect / OutcomeSet generalization across splice, readthrough, germline, and RNA evidence


iskandr:
## Background

PR #292 introduces \`SpliceOutcomeSet\`, a multi-outcome wrapper for splice-disrupting variants. The same "one DNA event → multiple plausible proteins" pattern appears in several adjacent workstreams:

- **Splice disruption** (#262, shipped in PR #292) — exon skipping / intron retention / cryptic splice / normal splicing.
- **RNA-level evidence** (#259) — enumerate plausible splice outcomes from DNA, narrow to observed isoforms using RNA-seq.
- **Germline-aware annotation** (#268, Sub-1 is #282) — the same somatic variant expressed against different germline backgrounds produces different haplotype-level outcomes.
- **Phasing / cis-trans composition** (#269) — two nearby variants on the same or different haplotypes produce different composite effects.
- **Future**: stop-loss with readthrough, alternative start codons, selenocysteine resolution (#42).

Each of these will want to wrap an underlying effect in a set of plausible outcomes. Committing to the \`SpliceOutcomeSet\` name / shape pigeonholes the abstraction into splicing.

## Scope

- Introduce a \`MultiOutcomeEffect\` protocol / base class (already added in PR #292's follow-up as a minimal marker: \`.candidates\`, \`.most_likely\`, \`.priority_class\`).
- Rename / refactor \`SpliceOutcomeSet\` to inherit from it cleanly and expose splice-specific fields (\`disrupted_signal_class\`, \`candidate_proteins\`, etc.) as extensions of the generic surface.
- Design the \`Candidate\` side analogously — \`SpliceCandidate\` becomes a subclass of a generic \`OutcomeCandidate\` carrying \`(outcome_label, plausibility, description, coding_effect, predicted_class_name)\`.
- Consider integrating with \`predict_variant_effects\` uniformly: a single \`multi_outcome=True\` (or \`outcome_sets=True\`) kwarg that wraps any multi-outcome-capable effect, of which splice is the first implementation.
- Downstream consumers (isovar, vaxrank, neoantigen pipelines) write \`isinstance(e, MultiOutcomeEffect)\` rather than \`isinstance(e, SpliceOutcomeSet)\`, so future wrappers don't churn their filtering code.

## Ordering

[...]

PR #310, opened 2026-04-15
[WIP #271 stage 3d] SequenceDiffEffectAnnotator scaffolding + SV design notes

iskandr:
**Draft / WIP — not for merge.** Skeleton for #309 so downstream reviewers (Vaxrank, Isovar) can comment on shape and extensibility before the algorithmic work starts.

> **User-facing documentation**: [docs/effect_annotation.md](docs/effect_annotation.md) is the canonical guide to how legacy, sequence-diff, splice outcomes, `MultiOutcomeEffect`, and the SV roadmap fit together. Start there for the big picture; this module docstring covers classifier-specific implementation detail.

## Shipping plan

| Order | Issue / PR | Scope |
|---|---|---|
| 1 | **#299 Part 1** (new PR, not filed yet) | Retrofit `ExonicSpliceSite` as `MultiOutcomeEffect`; `alternate_effect` becomes a derived property on both `ExonicSpliceSite` and `SpliceOutcomeSet`. Pure refactor, no behaviour change. |
| 2 | **This PR** (after #299 Part 1 lands) | Wire `apply_variant_to_transcript` + classifier + dual-dispatch against the unified shape. |
| 3 | **3e** (separate PR) | Flip default annotator to `sequence_diff` once the parity corpus passes. |
| parallel | **#305** | Rewrite `splice_outcomes.py` on `MutantTranscript`. Only needs stage 2's builder + 3d's classifier — can proceed independently of the default flip. |

## What's in THIS PR

- `varcode/annotators/sequence_diff.py` — class skeleton, classifier signature with the full decision table in the docstring, splice-adjacency gate stub with the dual-dispatch contract documented. All methods raise `NotImplementedError`. Not registered with the annotator registry.
- `tests/test_sequence_diff_parity.py` — parity harness skeleton with corpus composition spec, `pytest.mark.parity` marker, and INFO-logging rollout policy documented.
- `MutantTranscript` docstring extended with the SV extensibility roadmap (`reference_segments`, fusion / translocation / inversion handling).
- `docs/effect_annotation.md` — new user-facing guide linked from `docs/index.md`.
- Module docstring sections on: SV generalization, `MutantTranscript` ↔ `ProteinFragment` ownership boundary, possibility-set prior, recurrent-junction registry open question, dual-dispatch splice-adjacency contract.

## Feedback welcome on

1. **Classifier decision table** — anything missing? Edge cases worth calling out?
2. **Splice-adjacency gate** — resolved as dual-dispatch for exon-body, legacy-only for pure-intronic. See `_variant_is_splice_adjacent` docstring for the contract.
3. **SV extension shape** — does the `reference_segments` generalization match what Vaxrank / Isovar would consume? See module docstring + `docs/effect_annotation.md`.
4. **Corpus composition** — planned list in `test_sequence_diff_parity.py` docstring. What else belongs?
5. **Recurrent-junction registry** — most underspecified open question. Varcode-ships vs. plugin-consults? Not blocking 3d; likely spawns its own tracking issue when #252 starts.

Linked: #271 tracking, #304 staging, #309 implementation plan, #305 splice_outcomes rewrite, #299 MultiOutcomeEffect formalization, #252 / #257 / #259 / #297 — SV / fusion / RNA-evidence / ML-splice tracking.

### DIFF
--- a/docs/effect_annotation.md
+++ b/docs/effect_annotation.md
@@ -1,0 +1,291 @@
+# Effect annotation
+
+How varcode turns a variant into one or more `MutationEffect`
+objects, and how the pieces fit together across the basic case,
+splice-disrupting variants, and (eventually) structural variants.
+
+## The pipeline
+
+```
+Variant ─(Annotator)─> MutantTranscript(s) ─(Classifier)─> Effect(s)
+                              │
+                              └── wrapped in MultiOutcomeEffect
+                                   when >1 plausible outcome
+                                       │
+                                       └── narrowed by RNA evidence
+                                           (isovar / Exacto plug-ins)
+```
+
+Single DNA event can produce one or more plausible mutant
+proteins. varcode represents each concrete mutant as a
+`MutantTranscript`. When the DNA alone is ambiguous, the
+possibility set is wrapped in a `MultiOutcomeEffect`. RNA
+evidence (when available) narrows the set.
+
+## The four primitives
+
+| Primitive | What it represents | Module |
+|---|---|---|
+| `MutationEffect` (and subclasses) | One deterministic consequence: `Substitution`, `Silent`, `FrameShift`, `PrematureStop`, ... | `varcode.effects.effect_classes` |
+| `MutantTranscript` | One concrete mutant protein with edit provenance and the annotator that produced it | `varcode.mutant_transcript` |
+| `MultiOutcomeEffect` | Possibility set: a sequence of candidate outcomes ordered by a prior (most likely first) | `varcode.effects.effect_classes` |
+| `EffectAnnotator` | How a variant becomes effects or mutant transcripts | `varcode.annotators` |
+
+Everything in effect annotation is an implementation or consumer
+of one of those four.
+
+## Basic usage
+
+```python
+import varcode
+
+variants = varcode.load_maf("my_variants.maf")
+
+# Simplest path: get an EffectCollection.
+effects = variants.effects()
+effects.top_priority_effect()
+
+# Filter by transcript consequence.
+nonsilent = effects.drop_silent_and_noncoding()
+```
+
+`variants.effects()` calls the current default annotator
+(`"legacy"`) on every `(variant, transcript)` pair and returns
+an `EffectCollection`. Each element is a `MutationEffect`
+subclass — `Substitution`, `Silent`, `PrematureStop`, and so
+on.
+
+## Splice-disrupting variants: two representations
+
+When a variant sits in the canonical splice window (last 3
+exonic bases, first 3–6 intronic, canonical donor/acceptor),
+varcode recognizes it as splice-disrupting. Two ways the
+effect is expressed, at different richness levels.
+
+### Default: lightweight 2-outcome form
+
+```python
+variant = Variant("17", 43082575 - 5, "C", "T", "GRCh38")
+effect = variant.effect_on_transcript(transcript)
+# ExonicSpliceSite(...)
+#   .alternate_effect -> Substitution(...)  # if splicing proceeds
+```
+
+`ExonicSpliceSite` carries `alternate_effect`: the coding
+consequence that applies *if splicing still works*. Exactly
+two outcomes, represented as a primary effect + one
+alternate field. Cheap. Ships unconditionally.
+
+`SpliceDonor`, `SpliceAcceptor`, and `IntronicSpliceSite`
+don't expose `alternate_effect` today because the variant
+is intronic — there's no coding consequence to attach.
+
+### Opt-in: full possibility set
+
+```python
+effects = variant.effects(splice_outcomes=True)
+# SpliceOutcomeSet(...) replaces the splice effect
+#   .candidates ordered most-plausible-first:
+#     SpliceCandidate(NORMAL_SPLICING, plausibility=0.1,
+#                     coding_effect=Substitution(...))
+#     SpliceCandidate(EXON_SKIPPING, plausibility=0.5,
+#                     coding_effect=Deletion(...))
+#     SpliceCandidate(INTRON_RETENTION, plausibility=0.3)
+#     SpliceCandidate(CRYPTIC_DONOR, plausibility=0.1)
+```
+
+`SpliceOutcomeSet` replaces the splice effect with a set of
+candidate outcomes, each carrying a plausibility score
+(hand-tuned heuristic, not a probability) and — where
+computable from cDNA — a concrete `coding_effect`. The
+`NORMAL_SPLICING` candidate carries the same information as
+`alternate_effect` in the default form.
+
+When you opt in, `SpliceDonor` / `SpliceAcceptor` /
+`IntronicSpliceSite` also get wrapped, so every splice-
+disrupting variant produces a `SpliceOutcomeSet`.
+
+### Relationship between the two
+
+`SpliceOutcomeSet` is the N-outcome case of the same idea
+`alternate_effect` expresses with 2 outcomes. The
+`NORMAL_SPLICING` candidate is `alternate_effect` promoted to
+first-class; the other candidates express alternatives you
+don't see from the default form. Both are progressions of
+the same pattern:
+
+| # candidates | Class |
+|---|---|
+| 1 | plain `Substitution` / `Silent` / etc. — not wrapped |
+| 2 | `ExonicSpliceSite` with `alternate_effect` |
+| N (≥3) | `SpliceOutcomeSet` (opt-in via `splice_outcomes=True`) |
+
+After the planned `ExonicSpliceSite` retrofit in [#299][i299],
+all three will be `MultiOutcomeEffect` subclasses. Downstream
+code will be able to write `isinstance(e, MultiOutcomeEffect)`
+and iterate `.candidates` uniformly, regardless of richness
+level. `alternate_effect` becomes a derived property available
+on both `ExonicSpliceSite` (the 2-outcome case) and
+`SpliceOutcomeSet` (where it resolves to the
+`NORMAL_SPLICING` candidate's `coding_effect`). Same field,
+same meaning, works on both shapes — no parallel mechanism.
+
+### Variants in the CDS and in the splice window simultaneously
+
+A missense at the last exonic base is biologically *both* a
+splice effect (disrupts the splice signal) AND a protein-level
+coding effect (changes a residue). This is exactly what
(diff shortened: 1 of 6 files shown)


iskandr:
## Vaxrank-consumer feedback on the plan

Writing from vaxrank's perspective — we're downstream of both the variant→effect path (via `load_vcf_fast` + `variant.effects()`) and the mutant-protein-sequence path (currently via isovar, eventually possibly via `apply_variant_to_transcript` per #271's trajectory). Confirmed compat: vaxrank test suite passes clean on varcode 2.6.0, so the refactor series so far hasn't regressed anything visible downstream.

### On the classifier decision table

Shape is right — reduction to `(ref_delta, alt_delta)` via `trim_shared_flanking_strings` is the natural primitive, and the fall-through to `ComplexSubstitution` as the catch-all is the right escape hatch. Three things I'd pin in the corpus:

1. **`AlternateStartCodon` + downstream missense in the same variant.** The decision table reads the `AlternateStartCodon` test as "first codon non-ATG, proteins otherwise match." What if the variant *also* changes a non-initiator codon (two SNVs in the same codon on a phased allele, or an MNP)? That should fall through to `Substitution` / `ComplexSubstitution` — worth one regression test pinning "AlternateStartCodon is only taken when the only change is at codon 0."

2. **MT codon table.** #300 fixed a MT bug and the decision table reads as codon-table-agnostic (only compares protein strings), which is correct — but that means MT-specific classification correctness hinges entirely on `apply_variant_to_transcript` producing the right mutant protein string. Worth pinning at least one MT StopGain in the corpus since MT chromosomes use `TGA→W` not `TGA→stop`.

3. **StopLoss with NO 3'UTR readthrough extension.** The table reads `ref_delta ends at len(ref) and len(mut) > len(ref)` → StopLoss. What about a mutation that destroys the stop codon but happens to hit another stop codon a few residues downstream? The mutant protein would still be longer than ref (`len(mut) > len(ref)`), but the "extension" is bounded. That's still correctly `StopLoss` per the table, but the corpus should cover both "extension runs out of UTR" and "extension hits a downstream stop quickly" cases.

4. **Silent on pure-insertion of 3-residue multiple that exactly matches surrounding.** Legacy got this wrong pre-#201 — worth a corpus entry where the inserted residues are identical to flanking residues so `trim_shared_flanking_strings` could over-trim. E.g., inserting `AAA` between two existing `AAA` runs.

### On the splice-adjacency gate

"Within 3bp of an exon-intron boundary, delegate to legacy" is defensible near-term, but one worry from vaxrank's angle:

A splice-adjacent variant that *also* has a protein-level coding consequence (e.g. a missense at the last exonic base) is biologically *both* things. If the gate exclusively delegates to legacy, does the legacy path surface *only* the splice effect, or does it also produce the protein-level effect? Vaxrank's vaccine-peptide pipeline cares about the protein consequence; if a splice-adjacent missense gets classified as `ExonicSpliceSite` only and the protein change is invisible, we lose the neoantigen candidate silently.

Two options worth considering: (a) the gate delegates *additionally* to sequence-diff for the protein classification when the variant lies in the exon body, returning both effects; (b) the gate stays exclusive but documents explicitly that exon-body variants within 3bp of a boundary are legacy-only for both splice + protein classification. Either is fine — but this decision should be explicit in the docstring, not accidental.

### On the SV extension shape (`reference_segments`)

The "one DNA event → possibility set of plausible mutant proteins, narrowed by RNA evidence" framing matches exactly what neoantigen pipelines need to produce. Two concrete concerns for how this composes with downstream tools:

1. **Overlap with topiary's `ProteinFragment`.** Topiary 5.4.0 already ships `ProteinFragment` as the universal "mutant protein with source-type tag + target intervals" substrate (per openvax/topiary#121). For SVs, topiary expects downstream tools to produce `ProteinFragment`s — ideally one per candidate outcome, with `fragment_id` threading back to the genomic event. The `MutantTranscript` → `ProteinFragment` conversion should be a straight 1:1 mapping (transcript-level sequence becomes fragment sequence, provenance fields carry through). Worth sketching that conversion in the doc so downstreams know: "use `MutantTranscript` for transcript-level reasoning (splicing, UTR readthrough); convert to `ProteinFragment` at the prediction boundary." Otherwise we're going to see tools produce ad-hoc conversions that drift.

2. **Who owns the possibility-set narrowing?** The plan says "RNA evidence (#259) narrows the set at a later stage — for most DNA-only tools, the possibility set IS the answer." For vaxrank, this matters a lot: if varcode emits 50 candidate ORFs from a gene-to-intergenic translocation, and vaxrank hasn't been taught to prioritize across them, we'd silently produce 50 vaccine peptide sets. A caller-side *sort* order (most-probable-first by some DNA-only heuristic) would help even when no RNA evidence is available, so the top candidate is usable even if the tail is noise. Think about whether `MultiOutcomeEffect` should carry a prior over its outcomes, even a crude one (e.g., canonical-transcript-preferred for fusions).

3. **For fusions with known recurrent junctions:** possibility-set size 1 is the goal. But the mechanism for "known recurrent junction → single outcome" lives where? Is there a registry, or does the fusion annotator consult a lookup table? If it's the annotator's job, vaxrank/Exacto plugging in would need to know the registry's contract.

### On the corpus composition

The 500-variant plan is a good start. Suggested additions for vaxrank coverage:

- **Phased MNP where two SNVs hit the same codon** (tests the decision table's behavior when `ref_delta` and `alt_delta` are both length 1 but represent compound substitution).
- **Variant at the last exonic base that also changes the amino acid** (tests the splice-adjacency gate's protein-consequence handling — see above).
- **MT variants for at least PrematureStop and a stop-loss** (tests MT codon table path end-to-end).
- **A variant whose legacy effect_type was `ComplexSubstitution` pre-v2.0.0 but is now `Substitution` or vice versa** (parity delta worth pinning explicitly so we know it changed on purpose).

### On expected diffs

Worth pre-emptively documenting: the sequence-diff approach will implicitly *correct* any legacy offset-arithmetic bugs for cases where the protein-diff is trivially derivable. My guess for where to look:

- **Frameshifts that happen to realign to the original frame downstream** — legacy's offset-based logic may not notice; sequence-diff compares full strings and will classify correctly.
- **Indels in shared-prefix-with-mutant regions** — `trim_shared_flanking_strings` handles these cleanly; legacy's codon-by-codon walking can drift.
- **Multi-codon deletions where the deletion boundary is ambiguous** — sequence-diff canonicalizes via the shared-trim; legacy may report different offsets depending on walk direction.

Would recommend the parity harness log any classification deltas against legacy as INFO (not failures) during the initial corpus run, so expected diffs are visible and reviewable before the corpus is frozen as the parity contract.

### Smaller notes

- The `CORPUS = []` placeholder and module-level skip in the test file is the right shape for a WIP. Would suggest adding a `pytest.mark.parity` marker so the harness can be run selectively once populated.
- `NotImplementedError` stubs are clean. One minor: when the annotator is registered with the registry in #309, the `__repr__` should include "sequence_diff" by name so error messages like "`variant.effects(annotator='sequence_diff')`" are unambiguous to users.

Overall the plan is coherent; the SV extensibility section is the part most worth continuing to refine since it's the cross-cutting shape that downstream tools (vaxrank, topiary, Exacto, Isovar) will all consume. The "`MutantTranscript` → `ProteinFragment`" conversion boundary is the specific place to make sure the ownership is explicit.

iskandr:
Thanks — folded all five items directly into the scaffolding (commit 5673694 on this branch). Reads fresh now instead of buried in review comments.

### 1. Classifier edge cases → corpus

Pinned in the ``tests/test_sequence_diff_parity.py`` docstring as explicit planned additions: AlternateStartCodon + downstream missense (phased MNP / same-codon compound), MT PrematureStop + MT StopLoss (codon table end-to-end), StopLoss with bounded vs unbounded 3'UTR readthrough, pure 3-residue insertion matching flanks (the pre-#201 regression).

### 2. Splice-adjacency gate → dual-dispatch

Resolved in favor of **(a) dual-dispatch**. ``_variant_is_splice_adjacent`` docstring now makes this the contract:

- Exon-body variants in the splice window → legacy for splice classification, sequence-diff for protein-level classification attached as ``alternate_effect`` on the resulting ``ExonicSpliceSite``.
- Pure-intronic splice variants (inside the intron) → legacy-only since no protein-level diff exists to compute.

You're right that exclusive legacy delegation would silently drop the neoantigen-relevant protein consequence — the dual-dispatch avoids that.

### 3. SV ownership → explicit split + prior + registry

Three new sections in the module docstring:

- **MutantTranscript ↔ ProteinFragment split**: varcode owns transcript-level reasoning, topiary owns prediction-boundary framing, conversion is 1:1 at the boundary. Called out to prevent ad-hoc conversions drifting across tools.
- **Possibility-set prior**: ``MultiOutcomeEffect.most_likely`` returns the highest-prior candidate for DNA-only callers; RNA-evidence callers filter ``.candidates`` by observation support. Per-variant-class prior schemes outlined (splice uses existing plausibility tables; fusions prefer canonical transcripts; translocations rank by in-frame + ORF length + canonical involvement).
- **Recurrent-junction registry**: flagged the three open questions (varcode-ships vs annotator-consults, data source, common lookup interface) as explicitly out of 3d scope but documented so fusion support doesn't lock in an answer too early. Likely gets its own tracking issue once #252 starts.

### 4. Corpus additions

Added your four to the planned list in the test docstring: phased MNP (same codon), last-exonic-base missense (gate dual-dispatch), MT PrematureStop + StopLoss, cross-major-version effect_type drift.

### 5. Parity harness → INFO during corpus build

Rollout policy documented: during initial corpus build, deltas log as INFO via ``logging.info``. Each delta gets triaged — sequence-diff bug gets fixed; legacy bug sequence-diff corrects gets pinned in ``EXPECTED_DIFFS`` with an issue link. Only after every delta is triaged does the harness flip to ``assert``. Contract is **reviewed**, not assumed.

### Smaller notes addressed

- ``pytest.mark.parity`` registered in ``pyproject.toml`` — ``pytest -m parity`` selects the harness cleanly once populated.
- ``__repr__`` on ``SequenceDiffEffectAnnotator`` now includes the registered name so error messages map back to the ``annotator=`` kwarg string.

All files stay scaffolding — no classifier code ships in this commit. 606 tests pass + 1 parity skip, ruff clean. Diff is additive documentation only.

One item worth raising back: the recurrent-junction registry open question is the most underspecified thing in the design right now. If Vaxrank has a preference (varcode-ships vs plugin-consults), worth pinning now — harder to adjust once fusion annotators start depending on whatever interface we pick.

Issue #302, opened 2026-04-15
import varcode forces rpy2/R initialization via transitive PyVCF3 dependency


iskandr:
## Problem

Plain \`import varcode\` fully initializes rpy2 / R whenever rpy2 is installed in the environment.  Users see output like:

\`\`\`
INFO:rpy2.situation:cffi mode is CFFI_MODE.ANY
INFO:rpy2.situation:R home found: /opt/homebrew/Cellar/r/4.5.3/lib/R
WARNING:rpy2.rinterface_lib.openrlib:Error importing in API mode: ImportError(...libRblas.dylib...)
WARNING:rpy2.rinterface_lib.openrlib:Trying to import in ABI mode.
INFO:rpy2.rinterface_lib.embedded:Default options to initialize R: rpy2, --quiet, --no-save
INFO:rpy2.rinterface:R is already initialized. No need to initialize.
INFO:rpy2.rinterface_lib.embedded:Embedded R ended.
\`\`\`

R initialization is also surprisingly slow (~hundreds of ms on macOS), so this is a real import-time cost, not just noise.  \`rpy2\` is frequently installed as a transitive of unrelated scientific Python packages, so users who don't explicitly depend on it still hit the problem.

## Root cause

The chain is:

\`\`\`
varcode/__init__.py       # from .vcf import load_vcf, load_vcf_fast
  varcode/vcf.py          # import vcf as pyvcf  (PyVCF3)
    vcf/__init__.py       # from vcf.filters import Base as Filter
      vcf/filters.py
\`\`\`

\`vcf/filters.py\` (from PyVCF3) opens with an unconditional module-level probe:

\`\`\`python
try:
    from rpy2 import robjects
except:
    robjects = None
\`\`\`

So if \`rpy2\` is installed, every \`import varcode\` fully boots R — just to decide whether PyVCF3's optional R-based filter is available.

## Reproduction

\`\`\`bash
pip install varcode rpy2
python -c \"import varcode\"
\`\`\`

You'll see the rpy2/R init output above even though varcode never actually uses R.

## Proposed fixes (preference order)

1. **Upstream PyVCF3 patch (cleanest).** Lazy-import \`rpy2\` inside the filter class method that actually uses it, instead of at module top level.  PR against \`jamescasbon/PyVCF3\`.  PyVCF3 appears effectively unmaintained — PRs may stagnate.
[...]

PR #356, opened 2026-05-06
Drop runtime PyVCF3 dep: own VCF header/INFO/FORMAT parsing (#302)

iskandr:
## Why

PyVCF3 was the only reason `from varcode import load_vcf` dragged rpy2/R
into `sys.modules`. PyVCF3's `vcf/filters.py:1-4` does an unconditional
top-level `from rpy2 import robjects` — so any caller who resolved
`load_vcf` paid hundreds of ms of rpy2/R initialization, even when no
VCF was ever loaded. The lazy `__getattr__` shim in #346 deferred this
past `import varcode`, but only one indirection deep — `from varcode
import load_vcf` still triggered it.

What we actually used PyVCF3 for is small: read header metadata + sample
names, decode INFO and FORMAT cells (via the **private** `_parse_info` /
`_parse_samples` methods). Pandas already parses the row structure
(`varcode/vcf.py:read_vcf_into_dataframe`). So PyVCF3 was a heavyweight,
unmaintained-fork dependency for ~150 lines of work we can do ourselves.

## What changes

**New module: `varcode/vcf_parsing.py`** — `VCFHeader` dataclass with:

- `from_path(path)` / `from_lines(iterable)` — header parser
- `parse_info(info_str: str) -> dict` — INFO column decoder
- `parse_samples(sample_strings, format_string) -> OrderedDict` — FORMAT column decoder
- Reserved INFO/FORMAT key tables, mirroring PyVCF3 verbatim, for headers
  that don't declare common fields (DB, AF, GT, etc.)
- Quote-aware header splitter (so `Description="a, b, c"` doesn't get
  shredded by a naive `.split(",")`)
- Number=`.`/`A`/`G`/`R` encoded as `None` / `-1` / `-2` / `-3`, matching
  PyVCF3's `field_counts` exactly so a header round-trip yields the same
  `.number`.

**`varcode/vcf.py`**: drops `import vcf as pyvcf`, `import requests`,
`import zlib`. The `PyVCFReaderFromPathOrURL` wrapper (which had a dead
HTTP code path — `load_vcf` already downloads HTTP URLs to a temp file
above), `pyvcf_calls_to_sample_info_list`, and
`stream_gzip_decompress_lines` are deleted. `load_vcf` calls
`VCFHeader.from_path(path)` once and uses `header.parse_info` /
`header.parse_samples` as the parser callables it already accepted.

**`varcode/__init__.py`**: lazy `__getattr__` shim removed — `from
varcode.vcf import load_vcf` is no longer load-bearing for keeping rpy2
out of the import graph, because nothing in the import graph touches
PyVCF3 anymore.

**`requirements.txt`**: pyvcf3 removed. PyVCF3 is **not** a test
dependency either — its outputs were captured once and hardcoded as
literals in the test suite.

**`tests/test_lazy_vcf_import.py`**: deleted. The invariant it pinned
("`import varcode` doesn't pull rpy2") is now upheld unconditionally,
not via a workaround, and is covered by
`tests/test_vcf_parsing.py::TestNoRpy2OnImport`.

## How we know it's correct

PyVCF3's behaviour was the reference implementation; expected values
were captured from PyVCF3 and hardcoded as literals across three test
files:

- **`tests/test_vcf_parsing.py`** — parser internals and edge cases:
  quoted commas in `Description`, `Number=A` / `Number=R` / `Number=G`
  / `Number=.`, multi-value INFO with missing values, reserved-but-
  undeclared keys, INFO `.` (whole-field missing), `Flag` fields with
  and without `=`, FORMAT cells of `.` / empty / trailing-missing,
  undeclared FORMAT fields.
- **`tests/test_vcf_spec_examples.py`** — hand-decoded expectations
  against the canonical VCFv4.2 §1.1 trio and VCFv4.3 §5.4 SV examples
  from samtools/hts-specs.
- **`tests/test_real_world_callers.py`** — hand-verified expectations
  for fixtures from MuTect2 (verbatim from GATK), Strelka2, and VEP
  (format-faithful reconstructions; provenance documented in
  `tests/data/real_callers/README.md`).

Test results: **1014 passed, 2 skipped**. Ruff lint clean.

## Test plan

- [x] `./test.sh` — full suite green
- [x] `./lint.sh` — clean
- [x] `python -c "import sys; import varcode; print([m for m in sys.modules if m=='vcf' or m.startswith('vcf.') or m.startswith('rpy2')])"` → `[]`
- [x] `python -c "from varcode import load_vcf; vc = load_vcf('tests/data/mutect-example.vcf', genome='hg19'); print(len(vc))"` → 10
- [x] `python -m pytest tests/test_vcf_parsing.py -q` — 43 passed


iskandr:
## Follow-up: spec-example fixtures + dedicated tests + usage script

Per follow-up review feedback ("add lots of tests and examples with full but small VCF examples from public sources"), this commit adds:

**Two canonical fixtures** from samtools/hts-specs (MIT):
- `tests/data/spec_examples/vcf42_spec_trio.vcf` — VCFv4.2 §1.1 three-sample trio. Exercises `Number=A`/`Number=2`, Flag with comma-in-`Description`, multiallelic ALTs, microsatellite indel, `ALT="."` rows, phased+unphased GT, missing HQ values.
- `tests/data/spec_examples/vcf43_spec_sv.vcf` — VCFv4.3 §5.4 structural variant example. Exercises symbolic ALTs (`<DEL>`/`<INV>`), `END=`, `CIPOS` pairs, `IMPRECISE` flag, `SVLEN` with `Number=.`.

**Two layers of tests:**
1. The existing PyVCF3-oracle parametrized tests in `test_vcf_parsing.py` now glob recursively under `tests/data/` and pick these up automatically — every header field, every row's INFO, every row's per-sample data is compared element-by-element against PyVCF3.
2. New `tests/test_vcf_spec_examples.py` (35 tests) pins each fixture against **hand-verified expected values from the published spec**, not just PyVCF3 equality. So a regression here shows up as a diff against the textbook example, which is much more readable than a parametrized failure.

**Runnable example** at `examples/vcf_header_usage.py` — pretty-prints the trio fixture's header and decoded rows, demonstrating the public surface (`VCFHeader.from_path`, `parse_info`, `parse_samples`) for callers who want the typed dicts without going through `load_vcf`. Self-bootstraps `sys.path` so it runs from any cwd. Pinned by a smoke test.

**Counts**: 1072 passed, 8 skipped (was 1024). Ruff clean.

PR #396, opened 2026-07-08
PR: Fix protein_diff frameshift C-terminus truncation on a coincidental shared suffix

GeorgesBed:
## Summary

For a frameshift (or stop‑loss), the default `protein_diff` annotator returns a
`mutant_protein_sequence` that is **1–2 residues short at the novel C‑terminus** whenever the
novel tail coincidentally **ends with the same residue(s) as the reference protein's own
C‑terminus**. `Variant.effects()` (default annotator) is affected; `annotator="fast"` and
`predict_variant_effect_on_transcript()` are not.

## Root cause

`classify_from_protein_diff` (`varcode/effects/classify.py`) derives the frameshift's novel
tail from `trim_shared_flanking_strings(ref_protein, mut_protein)`, which trims a shared
**prefix _and_ a shared suffix**, then passes the suffix‑trimmed `alt_delta` as
`shifted_sequence`:

```python
ref_delta, alt_delta, prefix, suffix = trim_shared_flanking_strings(ref_protein, mut_protein)
...
return FrameShift(..., aa_mutation_start_offset=aa_offset, shifted_sequence=alt_delta)
```

Suffix trimming is correct for an **in‑frame** indel (reference and mutant genuinely share a
C‑terminal region), but a **reading‑frame change** produces a *novel* C‑terminus that runs to a
*new* stop codon — so any residue it happens to share with the reference protein's C‑terminus is
**coincidental** and must be kept.

## The fix

Use a **prefix‑only** trim (`trim_shared_prefix`) for the reading‑frame‑changing branches
(`FrameShift` / `StopLoss` / `FrameShiftTruncation`). The in‑frame branches are unchanged. The
`FrameShiftTruncation` guard becomes “no residues after the shared prefix” (`len(fs_shifted) == 0`),
which also fixes the extreme case where a 1‑residue novel tail coincidentally equal to the
reference's last residue was misclassified as a no‑novel‑sequence truncation.

## Verification

**Confirmed example — ATM p.F61fs, GRCh38 `chr11:g.108227882delT`, `ENST00000675843`:**

| path | before | after |
|------|--------|-------|
| `Variant.effects()` (default `protein_diff`) | `73 aa … FRKKQN` | `74 aa … FRKKQNV` |
| `Variant.effects(annotator="fast")` | `74 aa … FRKKQNV` | `74 aa … FRKKQNV` |
| `predict_variant_effect_on_transcript()` | `74 aa` | `74 aa` |

The true ORF ends `…N‑V‑[TGA]` (74 aa); ATM's reference protein also ends in `V`, so the shared
suffix `'V'` was being trimmed.

## Tests

Adds `tests/test_frameshift_cterminus_regression.py` (2 tests, Ensembl 115):

- `test_protein_diff_frameshift_keeps_cterminus_matching_reference_suffix`, ATM p.F61fs via
  `protein_diff` must be 74 aa ending `RKKQNV`.
- `test_fast_and_protein_diff_agree_on_frameshift_tail`, the two annotators must agree.

---


iskandr:
Merged as `dead346` and released in **6.0.2**. Thank you for the sharp catch and the thorough write-up — the root-cause analysis (shared-*suffix* trimming being wrong for a frameshift's novel C-terminus) made this easy to confirm and review. Verified end-to-end: on `main` the default annotator gave ATM p.F61fs as `73 aa …FRKKQN`, and now matches `fast` and the true ORF at `74 aa …FRKKQNV`; full test suite green. Filed #397 as the tracking issue (now closed). 🎉

Issue #321, opened 2026-04-17
Annotators disagree on aa_pos for deletions in repeat-residue stretches (HGVS 3'-rule)


iskandr:
## Summary

`FastEffectAnnotator` and `ProteinDiffEffectAnnotator` report **different amino-acid positions** for in-frame deletions that fall inside a repeated-residue stretch. Both agree on the effect *class* (`Deletion`) but the `aa_mutation_start_offset` differs by the length of the repeat, which means the `short_description` differs.

## Divergence

Deleting the first L codon (cDNA 429–431) in CFTR, where the surrounding protein is `…P L L L …` at positions 98–101:

- `fast` → `Deletion('p.L99del')`  — reports the *deleted codon's* position.
- `protein_diff` → `Deletion('p.L101del')` — reports the *HGVS-3'-canonical* position after whole-protein shared-flank trimming.

Pinned in `tests/test_annotator_divergence_scenarios.py::test_divergence_deletion_in_repeat_residue_stretch`.

## Analysis

Per HGVS nomenclature (§3.1, \"Recommendations for 'Description of DNA sequence variants'\"): *\"for all descriptions the most 3' position possible of the reference sequence is arbitrarily assigned to have been changed.\"* For a deletion inside a homopolymer residue run, this means the **last** residue of the run is canonically the one reported as deleted. Under that rule, `protein_diff`'s `p.L101del` is the canonical HGVS output.

### Why they differ

**Fast** (`varcode/effects/effect_prediction_coding_in_frame.py`) computes `aa_mutation_start_offset = ref_codon_start_offset + n_aa_shared + is_insertion` where the trimming runs only over the *codons touched by the edit* (`aa_ref = transcript.protein_sequence[ref_codon_start_offset:ref_codon_end_offset]`). For a single-codon deletion, `aa_ref = \"L\"` and `aa_alt = \"\"`, so no trimming happens and the offset is `ref_codon_start_offset = 99`.

**protein_diff** (`varcode/effects/classify.py::classify_from_protein_diff`) runs `trim_shared_flanking_strings(ref_protein, mut_protein)` over the *entire* pre- and post-edit protein. For `ref = \"…PLLL…\"` vs `mut = \"…PLL…\"`, the shared prefix extends two residues past the edit, pushing `aa_offset` from 99 to 101.

## Impact

[...]

Issue #318, opened 2026-04-17
protein_diff annotator: 3'UTR SNVs misclassified as Silent


iskandr:
## Summary

`ProteinDiffEffectAnnotator` classifies 3'UTR (and 5'UTR / Intronic) SNVs as `Silent` instead of passing through the location-based classification from `FastEffectAnnotator`.

## Divergence

For a 3'UTR SNV:

- `fast` → `ThreePrimeUTR` ✅
- `protein_diff` → `Silent` ❌

A test pinning this divergence lives in `tests/test_annotator_divergence_scenarios.py`.

## Root cause

In `varcode/annotators/protein_diff.py`, the annotator calls `fast_effect` early (for splice detection) but only defers to it for splice classes. When the mutant protein sequence matches the reference (which is true for UTR/intronic variants since they don't alter the CDS), the annotator falls through to classification and returns `Silent`, discarding the UTR/Intronic location information that `fast_effect` already carries.

## Suggested fix

After computing `fast_effect`, add a pass-through guard before the protein-diff path:

```python
from varcode.effects import ThreePrimeUTR, FivePrimeUTR, Intronic

if isinstance(fast_effect, (ThreePrimeUTR, FivePrimeUTR, Intronic)):
    return fast_effect
```

This is consistent with the existing splice pass-through pattern.

## Impact

Any user running `variant.effects()` under `protein_diff` on a VCF with UTR or intronic variants will see them collapse into `Silent`, losing the locational distinction that downstream tools (e.g. neoantigen pipelines, UTR variant annotators) rely on.

PR #398, opened 2026-07-08
Switch default annotator to fast (7.0.0); pin frameshift C-terminus parity

iskandr:
## Summary

Makes **`fast` the default effect annotator** again (replacing `protein_diff`), and pins the frameshift coincidental-shared-suffix class in the divergence suite. Major version bump to **7.0.0** because default behavior changes.

`Variant.effects()` / `VariantCollection.effects()` with no explicit `annotator=` now route through `fast`. `protein_diff` stays available via `annotator="protein_diff"` or `varcode.use_annotator("protein_diff")`, and remains the substrate the `MutantTranscript` / splice-outcome / germline machinery builds on.

## Why

`fast` is the more battle-tested path. During the `protein_diff` bring-up it was effectively the correctness oracle `protein_diff` was reconciled against (#318–#321, all cases where `protein_diff` was wrong and `fast` was right), and it has the cleaner bug history — most recently #396/#397, where the default `protein_diff` path dropped 1–2 residues off a frameshift's novel C-terminus while `fast` was correct.

For SNVs / indels / MNVs the two annotators are now fully reconciled (`test_protein_diff_parity.py`, `test_annotator_parity_adversarial.py`, `test_annotator_divergence_scenarios.py`), so nearly all output is unchanged; any residual divergence now resolves to `fast`'s classification.

## Changes

- `registry.py`: `_DEFAULT_NAME = "fast"` + updated rationale docstrings.
- Regression pins in `test_annotator_divergence_scenarios.py` for the frameshift coincidental-suffix class (#396/#397): CFTR `p.L127fs` (+ strand, tail `…GYAFSLL`) and BRCA1 `p.R71fs` (− strand, tail `…VNLLKSY`) must retain the full novel C-terminus under **both** annotators, plus a direct byte-parity check.
- Docs (`effect_annotation.md`), `fast.py` docstring, and `test_default_annotator_matches_active_configuration` updated to match. The codebase was already half-switched — `conftest.py`, `effect_annotation.md`, and `variant.py` already assumed `fast`.
- `CHANGELOG.md` + `version.py` → 7.0.0.

## Verification

- Full suite: **1231 passed / 5 skipped** (only the pre-existing, unrelated `test_filter_effects_by_transcript_expression` deselected — fails identically on `main`).
- `--annotator=protein_diff` CI parity mode: green.
- `ruff check`: clean.

## Note for reviewers

Making `fast` the default while `protein_diff` remains the roadmap substrate means the default path and the splice/germline path use different classifiers. They're reconciled today, so this is safe — but it makes the `fast`↔`protein_diff` parity suite load-bearing. Recommend making that suite (including the `--annotator=protein_diff` full-suite run) a required CI gate.

https://claude.ai/code/session_01VNtZRyKZ7u9jiMGPQEbx4c


iskandr:
Follow-up commit `a013ab2`: wired up `tests/test_protein_diff_parity.py`, which was a module-level skip with an empty corpus and a `pass`-body test (skipped with a stale '#309 hasn't landed' reason — a parity gate that guarded nothing).

The real harness now compares the **full `mutant_protein_sequence` + frameshift `shifted_sequence`**, not just class + `short_description`. That's the gap that let #396 ship: class *and* short_description matched under both annotators while `protein_diff` silently dropped a C-terminal residue. It sweeps CFTR (+) and BRCA1 (−) coding windows (590-entry corpus) plus the curated #396/#397 cases, and guards against the empty-corpus vacuous-pass. Runs by default now.

Suite is **1234 passed / 4 skipped** (the 4 remaining skips are data/release-dependent coordinate guards, unrelated to this change and identical on `main`).

Issue #403, opened 2026-09-11
CLI scripts fail to import: pkg_resources is missing from recent setuptools


iskandr:
`varcode/cli/effects_script.py` and `varcode/cli/genes_script.py` import `pkg_resources` to find their `logging.conf`:

```python
import pkg_resources
logging.config.fileConfig(pkg_resources.resource_filename(__name__, 'logging.conf'))
```

Recent setuptools releases no longer provide `pkg_resources` (seen with setuptools 83.0.0 on Python 3.12), so importing either module fails:

```
varcode/cli/effects_script.py:14: in <module>
    import pkg_resources
E   ModuleNotFoundError: No module named 'pkg_resources'
```

As a result `tests/test_cli_effects.py` and `tests/test_cli_genes.py` error at collection and `./test.sh` fails in such an environment, independent of any code change.

Suggested fix: resolve the file with the standard library instead, e.g. `importlib.resources.files(__package__) / "logging.conf"` (Python 3.9+), passed to `logging.config.fileConfig`.

PR #416, opened 2026-09-15
Refuse structural variants given to point-variant annotators

### LINKED ISSUE #412: EffectAnnotator.supports is never checked; UnsupportedVariantError is never raised
The `EffectAnnotator` protocol docstring (`varcode/annotators/__init__.py:61-65`) says an annotator's `supports` set is enforced: "Callers that hand the annotator a variant outside this set get a clear `UnsupportedVariantError` rather than silently wrong output."

That doesn't happen anywhere:

- No code checks `.supports`. Nothing raises `UnsupportedVariantError` (`registry.py:35`) either, and it's exported from `varcode` and listed in `docs/api.md`.
- `effect_prediction.py:110-118` says so directly: "that flag is metadata — nobody dispatches on it". It only routes SVs to the SV annotator when `annotator is None`. With an explicit `annotator="fast"` or `"protein_diff"`, an SV runs through point-variant offset arithmetic on the placeholder `ref="N"`/`alt="A"`. That comment calls the output "nonsense".
- `tests/test_annotators.py:143-155` only checks that the exception subclasses `ValueError`, and notes "No code throws it yet".

The fix needs a decision:

1. Enforce it: check the variant's kind against `annotator.supports` in `predict_variant_effects` and raise `UnsupportedVariantError`. That needs a `variant_kind` helper that returns `snv`/`indel`/`mnv` or the SV type.
2. Drop it: remove `supports` from the protocol, delete `UnsupportedVariantError`, and fix the docstring.

I'd go with (1), since the SV routing special case already covers half of it.

iskandr:
Fixes #412.

## Bug
The point-variant path never reads `sv_type`. Given an SV, it works from the SV's span and its placeholder `N>A` allele. On `main`, with CFTR (`ENST00000003084`) and the same span for DEL, DUP and INV:

| SV | `annotator="fast"` | default (`structural_variant`) |
|---|---|---|
| BND CFTR→BRCA1 | `Intronic` | `GeneFusion` |
| DUP across exons 1–3 | `ExonLoss` | `LargeDuplication` |
| INV across exons 1–3 | `ExonLoss` | `Inversion` |
| DEL inside one exon | raises `ReferenceMismatchError` | `LargeDeletion` |
| INS in an exon | raises `ReferenceMismatchError` | `LargeDuplication` |
| DEL across exons 1–3 | `ExonLoss` (3 exons) | `LargeDeletion` (3 exons) |

The last row is essentially the same answer. It's the one case where span-based classification works. The others miss a fusion, turn a gain into a loss, or crash.

Every annotator declares a `supports` set, and `UnsupportedVariantError` is exported and documented. Nothing checked either one.

## Change
- **The check.** `predict_variant_effects` compares the variant's kind with the resolved annotator's `supports` set and raises `UnsupportedVariantError` on a mismatch. `Variant.effects` and `VariantCollection.effects` both go through it. It raises even when `raise_on_error=False`, because an unsupported variant is a caller error, not bad data.
  - A structural variant's kind is its `sv_type`. Point variants are `snv`, `indel` or `mnv`, with complex substitutions counted as `mnv`.
  - The structural check comes first, because `StructuralVariant.is_snv` is `True`.
- **Default routing is unchanged.** With `annotator=None`, SVs still go to `structural_variant`, including inside `use_annotator("protein_diff")`.
- **The structural annotator refuses point variants too.** `annotator="structural_variant"` on a point variant also raises.
- **Duck-typed annotators.** An annotator with no `supports` attribute isn't checked, so third-party annotators keep working.
- **Direct calls.** `annotate_on_transcript` itself is still unchecked. Checking every transcript would slow the fast path, and every public entry point already goes through `predict_variant_effects`.

## Behavior change
This reverses a deliberate choice. `test_vcf_loader_sv_explicit_protein_diff_passes_through` required explicit overrides to pass through for parity testing, and that test now expects the error. After this PR, `vc.effects(annotator="fast")` on a collection containing SVs raises. Before, it returned wrong effects for the SVs.

Nothing in vaxrank, topiary or isovar passes `annotator=`. Bumped to 7.2.0 for the behavior change, and added a CHANGELOG entry and a paragraph in `docs/effect_annotation.md`.

## Testing
- `tests/test_annotator_supports.py` (new, 16 tests). 7 of them fail on `main`.
- `./lint.sh` passes.
- `./test.sh`: 1293 passed, 0 failed.
  - The only errors are in `tests/test_cli_effects.py` and `tests/test_cli_genes.py`. They can't import `pkg_resources` in my local environment (#403); CI on `main` is green.

## Merging with #409
#409 edits docstrings in `annotators/__init__.py`, `registry.py` and `tests/test_annotators.py`, but in different hunks, so the two should merge cleanly.


iskandr:
Superseded by the user-approved return-value contract in merged PR #428: one unified default, partial experimental annotators return NotImplemented, and public APIs expose Unresolved instead of enforcing static supports sets.

Issue #414, opened 2026-09-14
deploy.sh doesn't bump, tag, or guard as AGENTS.md says; no tags for 2.0.0–5.0.6


iskandr:
`AGENTS.md` describes a `deploy.sh` that varcode doesn't have:

- `AGENTS.md:10`: "`deploy.sh <version>` handles the bump + commit + push"
- `AGENTS.md:25`: "gates on lint + test, refuses to run off `main`/`master`, and refuses a dirty tree"
- `AGENTS.md:32`: "lint → test → optional version bump → build → twine upload → tag → push"

The real `deploy.sh` is seven lines: `./lint.sh && ./test.sh`, `python3 -m build`, `twine upload dist/*`. It takes no arguments, doesn't check the branch or working tree, and doesn't tag or push. `RELEASING.md` gives a third, older version of the process: bump `version.py` by hand, merge into `master`, run `deploy.sh`.

You can see the result in the tags. PyPI has 2.0.0 through 5.0.6 (54 releases) with no git tags. The only tags since 0.6.1 are v6.0.0, v6.0.1 and v7.0.0, which were probably created by hand. The CHANGELOG also links to `tree/v2.x` tags that don't exist (#411).

isovar and vaxrank already have the deploy script AGENTS.md describes (`./deploy.sh [--dry-run] [version]`, with bump, guards, tag and push).

Proposal:
1. Port the isovar `deploy.sh` to varcode.
2. Delete `RELEASING.md`, or cut it to "run `./deploy.sh <version>` from a clean main".
3. Optionally backfill `vX.Y.Z` tags for 2.0.0–5.0.6 at the commits that bumped `varcode/version.py`.

PR #428, opened 2026-09-15
Unify default annotation and let experimental annotators decline inputs

### LINKED ISSUE #412: EffectAnnotator.supports is never checked; UnsupportedVariantError is never raised

### LINKED ISSUE #427: Annotator registry test leaks protein_diff as the process default
tests/test_annotators.py::test_set_default_annotator_swaps_registry_default unconditionally restores protein_diff in its finally block, although the shipped default is fast. Without --annotator scope isolation, subsequent tests run with the wrong default and can mask default-routing regressions. Restore the prior default rather than a hard-coded name. Found while implementing the return-value contract for #412.

iskandr:
## Summary

One built-in default, still named fast for compatibility, now owns point-edit and structural-variant routing. Explicit fast selection and direct transcript calls use the same route as effects(). Structural implementation stays in its own module.

Partial annotators require only name and annotate_on_transcript. They return NotImplemented for individual unsupported inputs instead of maintaining supports lists. Public prediction APIs return Unresolved(mechanism="unsupported_annotation", reason=...) and retain the selected annotator provenance, with no implicit fallback. None remains an invalid plugin result; genuine exceptions keep the existing error policy.

protein_diff and realized remain experimental options. The default biological model and ranking are not promoted to the experimental realized model. Germline context uses an explicit optional context hook; the default preserves its point-edit patient-baseline machinery and declines unsupported SV/germline composition. The structural-only entry point and UnsupportedVariantError import remain for compatibility.

## Compatibility

Version 8.0.0: removed supports metadata and consistent scoped/direct annotator selection change the public contract. Shared protein-comparison helpers and point-edit parity tests remain. CI compares optional implementations only on their shared domain; the full suite separately tests unsupported inputs.

## Verification

- ./lint.sh passed
- TEST_SH_MAX=4 ./test.sh: 1366 passed, no skips
- TEST_SH_MAX=4 COVERAGE_FILE=.coverage.parity ./test.sh --annotator=protein_diff -m parity: 3 corpus tests passed
- New regression tests cover partial plugins, all six SV kinds declined by protein_diff, context-specific refusal, mixed collections, JSON round-trip, provenance, actual errors, and internal germline selection isolation.

Closes #412. Closes #427. Supersedes the capability-list enforcement design in #416.

The short implementation spec is docs/annotator_contract.md. Release uses the existing deploy script from clean main; its already-filed guard/version/tag limitations remain tracked in #414.


iskandr:
Released as varcode 8.0.0: https://pypi.org/project/varcode/8.0.0/ . The deployment script passed lint and all 1366 tests from clean main at 50eafde994e1af53e3bbeb34d2d081de304778bd, then uploaded both distributions. Public PyPI SHA-256 hashes match the local wheel and sdist. The v8.0.0 tag points to that merge commit. PR #416 is closed as superseded.

PR #429, opened 2026-09-15
Fix CLI startup without pkg_resources (8.0.2)

### LINKED ISSUE #403: CLI scripts fail to import: pkg_resources is missing from recent setuptools

iskandr:
## Fix

`varcode` and `varcode-genes` fail on import when setuptools no longer provides
`pkg_resources`. Resolve their packaged `logging.conf` with the standard library
instead: `importlib.resources.files(__package__)` plus `as_file()` so zipped
packages also work. `logging.conf` is already included in package data.

Fixes #403. Patch release: **8.0.2** (8.0.1 has already shipped).

The original CLI implementation is retained; the release follow-up adds two
subprocess regression tests which block `pkg_resources`, plus the changelog.
No annotation semantics or unrelated CLI error-handling changes are included.
The breaking structural/default cleanup stays separate in draft #439.

## Verification

- Reproduced from PyPI 8.0.1 in a fresh Python 3.11 environment with setuptools
  84.0.0 and no `pkg_resources`: both installed commands exit 1 with
  `ModuleNotFoundError` on `--help`.
- Installed the built 8.0.2 wheel into that environment: both installed commands
  exit 0 and print their help, still without `pkg_resources`.
- `./lint.sh`: passed.
- Targeted CLI tests: **4 passed**.
- Full local suite in the fresh environment: **1,372 passed**, no skips.
- Both CLI modules also passed `--help` when imported directly from the zipped
  candidate wheel (exercises `importlib.resources.as_file`).
- GitHub tests on Python 3.9/3.10/3.11 and docs: green at
  `79cec3a9a65a0b482af2a19ca6e1e1e0b49ff0db`.

After merge: run the deployment script from clean main, verify public artifact
hashes, then test both installed commands in a second fresh environment using
only the published PyPI package before calling the release finished.

Original implementation session:


iskandr:
Release coordination: #430 has merged the transcript_model naming change with version 8.0.1 and is being deployed now. Please update this CLI PR to the next unused version before merging/releasing; its body already anticipates this case. No CLI changes were included in #430.

iskandr:
**Verdict:** merge after fixes. The fix is right for the install shapes that matter, but the version bump has gone dead since the PR was opened, and one install shape regresses relative to `pkg_resources`.

## Verified

Re-ran what the PR body claims, on `origin/cli-importlib-logging-conf` (a8a57ca):

- The bug is real on current `main` (9113af0): `pytest tests/test_cli_effects.py tests/test_cli_genes.py` → 2 collection errors, `ModuleNotFoundError: No module named 'pkg_resources'`. Root cause confirmed: `setuptools` is not in `requirements.txt`, so nothing pulls `pkg_resources` in.
- On the branch: `./lint.sh` clean; `pytest tests/test_cli_effects.py tests/test_cli_genes.py -q` → 2 passed.
- `python -m build` — both the wheel and the sdist contain `varcode/cli/logging.conf`, so `[tool.setuptools.package-data]` is doing its job and `MANIFEST.in` needs nothing.
- Editable install (this env uses setuptools' strict `__editable___varcode_8_0_0_finder` MetaPathFinder, not a plain `.pth`): `importlib.resources.files("varcode.cli")` returns a real `pathlib.PosixPath` → `fileConfig` works.
- Unpacked-wheel install: `varcode --help` prints usage and exits 0.
- API floor matches `requires-python = ">=3.9"`: `importlib.resources.files()` is 3.9+, and the 3.9 CI job is green.
- No `pkg_resources` references left anywhere in the tree.

## Problems

**1. `varcode/version.py` — the bump is now a no-op, and the PR would ship unversioned.**

`main` moved while this was open: #430 ("Rename experimental realized annotator to `transcript_model`") took 8.0.1, wrote a `CHANGELOG` entry for v8.0.1, and 8.0.1 is already the latest release on PyPI. This branch also sets 8.0.1. Because both sides make the *identical* edit to the same line, git merges it without a conflict:

```
$ git merge-tree --write-tree origin/main origin/cli-importlib-logging-conf
86f812e...
$ git show 86f812e:varcode/version.py
__version__ = "8.0.1"
```

So merging this PR bumps nothing (AGENTS.md rule 2), the CLI fix gets no release of its own, and the 8.0.1 already on PyPI is the transcript_model release — which still cannot run `varcode --help`. This needs `main` merged in and the version moved to 8.0.2, with #431 shifted to 8.0.3.

**2. `varcode/cli/effects_script.py:21`, `varcode/cli/genes_script.py:22` — `str(resources.files(...))` breaks under zip import.**

`resources.files()` returns a `zipfile.Path` when the package is zip-imported (zipapp, pex, shiv, a zipped egg, a wheel put directly on `sys.path`). `str()` of that is not a path `open()` can use, so `fileConfig` raises at import — the same failure mode this PR is fixing, just in a different environment. `pkg_resources.resource_filename` did not have this problem: its `ZipProvider` extracted the resource to a cache directory first.

Repro, wheel on `sys.path`:

```
$ PYTHONPATH=dist/varcode-8.0.1-py3-none-any.whl:... python -c "import varcode.cli.effects_script"
FileNotFoundError: .../varcode-8.0.1-py3-none-any.whl/varcode/cli/logging.conf doesn't exist
```

The supported spelling, also 3.9+, and verified to fix that repro:

```python
with resources.as_file(resources.files(__package__) / "logging.conf") as path:
    logging.config.fileConfig(path)
```

## Nits

- `__package__` works here, but `__spec__.parent` is the documented accessor and doesn't depend on `__package__` being populated.
- Unrelated to this PR, but now that both test modules import these scripts: `logging.config.fileConfig` runs at import time with its default `disable_existing_loggers=True`, so importing `varcode.cli.effects_script` silently disables loggers configured earlier in the process.
- Also pre-existing: `print_version_info()` runs before `parse_args`, so `varcode --help` prints the version block above the usage message.

iskandr:
Taking this through the requested 8.0.2 hotfix release now. I have fetched head ef1f347 and will add a regression test that blocks pkg_resources, record the changelog, and verify both console entry points from the built wheel and fresh PyPI installs before/after publishing. PR #439 remains a separate 9.0.0 draft; #431 is not being merged with this hotfix.

iskandr:
Released and verified as 8.0.2: https://pypi.org/project/varcode/8.0.2/. The deployment script passed lint and all 1,372 tests on clean main. Public wheel/sdist SHA-256 hashes match the release artifacts; tag v8.0.2 points to merge commit 3f6e9f773b6672b52a6b9d45007ef91e7bedf8f1. A second fresh Python 3.11 environment installed varcode==8.0.2 directly from PyPI with setuptools 84.0.0 and no pkg_resources: both varcode --help and varcode-genes --help exited 0. The same commands on a fresh PyPI 8.0.1 install failed with ModuleNotFoundError, confirming the regression and fix.

PR #431, opened 2026-09-15
CLI: one-line errors for common mistakes (8.0.2)

iskandr:
Stacked on #429, which is needed for the CLI to import at all. The base is `cli-importlib-logging-conf`; retarget to `main` once #429 merges.

Common mistakes now give one line and exit 1 instead of a 15–25 line traceback. Only `varcode/cli/` and the version change.

| Command | Before | After |
|---|---|---|
| `varcode` | traceback, `ValueError: No variants loaded ...` | `varcode: error: No variants loaded (use --maf, --vcf, --variant, or --json-variants options)` |
| `varcode --vcf /nonexistent/x.vcf` | traceback, `FileNotFoundError` | `varcode: error: [Errno 2] No such file or directory: '/nonexistent/x.vcf'` |
| `varcode --variant 12 25245350 C T` | traceback | `varcode: error: --genome must be specified when using --variant` |
| `varcode --genome GRCh38 --variant 12 abc C T` | traceback, `invalid literal for int() with base 10: 'abc'` | `varcode: error: --variant position must be an integer, got 'abc'` |
| `varcode --genome hg99 ...` | three chained `KeyError` tracebacks | `varcode: error: Failed to infer genome assembly name for hg99` |
| `varcode --genome GRCh38 --variant chrZ 100 A T` | traceback | `varcode: error: Invalid contig name 'chrZ' for reference 'GRCh38'` |
| `varcode --genome NCBIM37 ...` (not installed) | traceback | `varcode: error: GTF database needs to be created, run: pyensembl install --release 67 --species mus_musculus` |
| `varcode --vcf tests/data/strelka-example.vcf` | traceback | `varcode: error: Invalid contig name 'chrUn_gl000216' for reference 'GRCh37'` |
| `varcode ... --output-csv /nonexistent/dir/out.csv` | annotates everything, then `OSError` traceback | fails before loading: `varcode: error: output directory does not exist: /nonexistent/dir` |
| `varcode-genes --maf /nonexistent/x.maf` | traceback | `varcode-genes: error: [Errno 2] No such file or directory: '/nonexistent/x.maf'` |

Implementation: a small `exit_on_user_error(arg_parser)` context manager in `variant_args.py` turns `ValueError`/`OSError` into `parser.exit(1, ...)`. That's broad on purpose: varcode signals every user-input problem with `ValueError` (including `ReferenceMismatchError` and JSON decode errors). An unexpected `ValueError` from an internal bug would also lose its traceback. If that matters, a `--debug` flag could re-raise.

Not changed here, filed separately: the success-path output is still a logger line holding the `EffectCollection` repr. Some messages are still worth rewording (the reference-mismatch text shows transcript-strand bases; `chrUn_gl000216` shouldn't abort the whole file).

Bumps version to 8.0.2.

Tests: `./lint.sh`; `python -m pytest tests/test_cli_effects.py tests/test_cli_genes.py -n 0 -q` gives 8 passed. The 6 new cases fail without the `varcode/cli` changes.


iskandr:
**Verdict:** merge after fixes. The change does what it says and the tests are honest, but the broad `ValueError` catch needs an escape hatch, two user mistakes still traceback or still fail late, and the stacking needs attention.

Reviewed as the diff against `cli-importlib-logging-conf` (#429), i.e. 05bbfd1 vs a8a57ca.

## Verified

- Every row of the table in the PR body reproduces exactly as written, with exit 1 and a single line on stderr: no arguments, `--vcf /nonexistent/x.vcf`, `--variant` without `--genome`, `--variant 12 abc C T`, `--genome hg99`, `--variant chrZ 100 A T`, `--vcf tests/data/strelka-example.vcf`, `--output-csv /nonexistent/dir/out.csv`, and `varcode-genes --maf /nonexistent/x.maf`.
- `./lint.sh` clean; `pytest tests/test_cli_effects.py tests/test_cli_genes.py -q` → 8 passed.
- The 6 new cases really do depend on the fix. With `varcode/cli/` taken from #429 and the tests from this branch: 6 failed, 2 passed — each of the 6 fails with the raw exception (`FileNotFoundError`, `ValueError`, `OSError`) instead of `SystemExit`.
- `--output-csv` really is checked first: the `/nonexistent/dir` case exits before touching pyensembl, which is also why that test needs no genome data.
- Success paths unchanged: `varcode --genome GRCh37 --variant 12 25398284 C T` → 4 KRAS `p.G12D` effects, exit 0; `varcode-genes --maf tests/data/ov.wustle.subset5.maf` → table, exit 0; `--output-csv` writes the file.
- Extra error cases not in the table also come out as one line, exit 1: malformed `--json-variants`, an empty VCF, `varcode-genes --output-csv /nonexistent/dir/o.csv`.
- No downstream breakage from `parse_position`: topiary, isovar and vaxrank all call `variant_collection_from_args` / `make_variants_parser` / `add_variant_args`, and none of them are affected — a non-integer position raised `ValueError` before this PR too, just with a worse message.

## Problems

**1. `varcode/cli/variant_args.py:115` — the `ValueError` catch turns varcode's own internal invariants into "user errors".**

varcode raises `ValueError` from 81 sites across 27 non-CLI modules, and a large share of those are assertions about its own state rather than anything the user typed. For example:

- `varcode/genomic_layout.py:584` — `"Breakend retained an empty layout"`
- `varcode/splice_outcomes.py:1419` — `"Exon %s not found in transcript %s"`
- `varcode/effects/effect_ordering.py:627` — `"List of effects cannot be empty"`

Any of those now reaches the user as a bare line with no stack, indistinguishable from a typo:

```
$ varcode --genome GRCh37 --variant 12 25398284 C T     # with to_dataframe patched to raise that ValueError
varcode: error: Exon ENSE00000000001 not found in transcript ENST00000256078
```

The PR body raises this itself and suggests `--debug`. It matters because bug reports will now arrive as a single sentence. Narrowing the `with` block is not an option — `Invalid contig name` is raised lazily during `variants.effects()`, not during loading — so it needs an explicit opt-out. The cheapest version adds no flag and no help text: re-raise when `VARCODE_DEBUG` is set in the environment.

**2. `varcode/cli/variant_args.py:129` — an out-of-range `--variant` position still tracebacks.**

`parse_position` only rejects non-integers. A position that parses but doesn't fit in the Ensembl database's SQLite `INTEGER` gets through and blows up 28 lines deep:

```
$ varcode --genome GRCh37 --variant 12 99999999999999999999 C T
  ... 28 lines ...
OverflowError: Python int too large to convert to SQLite INTEGER
```

`OverflowError` is neither `ValueError` nor `OSError`, so `exit_on_user_error` doesn't see it. This is exactly the class of mistake (a mis-pasted coordinate) the PR is about.

**3. `varcode/cli/variant_args.py:119` — `check_output_directory` only checks the parent, so one output mistake still fails after the whole run.**

```
$ varcode --genome GRCh37 --variant 12 25398284 C T --output-csv /tmp
  ... full annotation runs ...
varcode: error: [Errno 21] Is a directory: '/tmp'
```

`os.path.dirname("/tmp")` is `/`, which exists, so the pre-flight passes. The message is fine; the timing is the thing this PR set out to fix.

**4. "One line" isn't uniform: `--maf` given a non-MAF file prints three lines with a pandas repr in them.**

```
$ varcode --maf tests/data/somatic_hg19_14muts.vcf
varcode: error: Too few columns in MAF file tests/data/somatic_hg19_14muts.vcf, expected 19 but got  10 : Index(['chr1', '53513530', '.', 'A', 'C', '..1', '..2', 'GE=SCP2;EG=6342',
       'GT', '0/1'],
      dtype='object')
```

The message is built in `varcode/maf.py:83-87`, outside this PR's blast radius, so I've left it — but it's worth a follow-up, along with the stray double space in `"got  %d"`.

**5. Stacking / retarget.** Base is `cli-importlib-logging-conf`. GitHub only auto-retargets to `main` if #429 is merged *and* its branch is deleted; merged-but-kept leaves this PR merging into a stale branch. Separately, #429's 8.0.1 now collides with `main` (see my comment there) — once that's rebased onto current `main` at 8.0.2, this one needs 8.0.3.

## Nits

- `varcode/cli/__init__.py` re-exports the CLI helpers under an explicit `__all__`, and the three new helpers aren't added. Fine if they're deliberately private, but the module now mixes two conventions.
- `--variant 12 0 C T` is accepted and reported as `Intergenic(... start=0 ...)`. varcode coordinates are 1-based, so 0 isn't a legal locus. Pre-existing, and `parse_position` is the obvious place to reject it, but it's a behavior change so I've left it to you.
- `exit_on_user_error`'s docstring lists "unwritable output path" among what it reports. A permission-denied output path is caught, but only after the whole annotation has run.

iskandr:
Release coordination: #429 is now merged and published as 8.0.2, including the pkg_resources startup fix and clean-install regression tests. Please rebase this separate error-handling PR onto current main and choose the next unused version before release; do not reuse 8.0.2. The 9.0.0 structural cleanup remains a draft in #439.

PR #439, opened 2026-09-15
Consolidate structural prediction into the default annotator (9.0.0)

iskandr:
## Scope / design

Finish the structural/default consolidation without adding an outside router:

- `fast` calls plain internal structural prediction helpers in
  `varcode.effects.structural`, instead of instantiating another annotator.
- The experimental `transcript_model` calls the same helpers for its existing
  breakend/fusion path. It remains experimental; `fast` remains the default.
- Keep the registry only for explicit annotator selection. No capability lists,
  automatic fallback, or changes to the `NotImplemented` / `Unresolved` contract.
- Remove the old public APIs rather than keeping compatibility shims, as
  explicitly requested. Version bump: **9.0.0**.

The user requested completion of the merges/releases before the osteosarc
phasing figure analysis. #440 has shipped as 8.0.3 and refreshed CI is green.

Rebased onto the **8.0.3 dataset-specific contig fix** (#440,
`24cbe2e935d29f958baec7d7a9ed0c8029be4ce4`), which also contains the released
8.0.2 CLI hotfix. The rebase changes only version/changelog context in this
cleanup patch; no cleanup logic changed.

## Breaking API removals / migration

| Removed | Replacement |
| --- | --- |
| `StructuralVariantAnnotator`, including its module | `FastEffectAnnotator` for direct raw annotation, or normal variant effect APIs |
| `annotator="structural_variant"` | Omit `annotator`, or select `"fast"` |
| `UnsupportedVariantError` from package/annotators/registry | Partial plugins return `NotImplemented`; public APIs expose `Unresolved` |

Pickles referencing the removed module/classes require varcode 8 to read before
migration. This PR does not remove the recent `realized` compatibility aliases.

## Behavior preserved

Structural classification, affected/junction coordinates, cDNA/protein building,
candidate order/evidence, and historical structural-builder provenance are
unchanged. Collections continue to record the selected annotator's provenance.
All 33 extracted functions/classes were checked against the original ASTs,
normalizing only method-to-function plumbing, import paths, docstrings and the
warning's obsolete class-name wording. The SV dispatch mapping is unchanged.

Existing structural, fusion and candidate tests now exercise the unified
default. New regression tests block imports of the retired module, cover every
supported SV kind with/without an assembly, compare default/helper results,
and verify removal of the public APIs and registry entry.

## Explicit follow-ups, not part of this PR

- **#437**: `VariantCollection.effects()` constructs joint haplotype effects
  outside the selected annotator. Filed with a reproducer: a plugin declining
[...]

PR #440, opened 2026-09-16
Fix dataset-specific contig validation (8.0.3)

### LINKED ISSUE #402: Contig validation cache conflates Genome datasets sharing an assembly name
## Reproduction and impact

Found while adding offline original-RNA protein validation in openvax/isovar#219. Varcode **5.0.6** caches valid contigs in `Variant._reference_name_to_valid_contig_names`, keyed only by `reference_name` (e.g. GRCh38), not the actual Genome/annotation dataset.

Loading a six-transcript GRCh38 annotation subset first caches only its six chromosomes. A subsequent full Ensembl GRCh38 query at TP53/chr17 is incorrectly rejected as an invalid chromosome. Order reverses the behavior: full-first permits chromosomes absent from the subset. Different releases/custom annotations sharing an assembly are also affected.

Minimal no-download reproduction of the affected helper:

```python
from types import SimpleNamespace
from varcode import Variant

Variant._reference_name_to_valid_contig_names.clear()
for label, contigs in [('subset', ['1']), ('full', ['1', '17'])]:
    v = SimpleNamespace(
        reference_name='GRCh38', contig=contigs[-1],
        genome=SimpleNamespace(contigs=lambda: contigs),
        _reference_name_to_valid_contig_names=Variant._reference_name_to_valid_contig_names)
    Variant._check_that_genome_has_contig(v)
    print(label, 'accepted')
# subset accepted
# ValueError: Invalid contig name '17' for reference 'GRCh38'
```

Integration reproduction with the PR's original six-transcript subset: run `tests/test_osteosarc_proteins.py` before `tests/test_variant_orf.py::test_protein_sequence_creator_translates_coding_deletion` with a subset Genome named GRCh38. The latter previously passing test fails. Its `variant.effects(raise_on_error=False)` also returns a plain `[]` for this error, so Isovar's subsequent `clone_with_new_elements` call raises AttributeError; the underlying problem is the incorrect chromosome cache.

## Acceptance

- Scope contig validation to the actual Genome/annotation dataset (or query its own cached `contigs()` directly); never allow one dataset to poison another sharing an assembly name.
- Tests for subset-first and full-first, two annotation versions on the same assembly, invalid-contig rejection, and repeated valid queries.
- Preserve consistent EffectCollection/error behavior for failed effect prediction; a swallowed contig failure must not unexpectedly change collection type.
- Downstream regression: a small offline dataset followed by a full-GRCh38 TP53 protein translation in the same process succeeds.

For the Isovar pilot, the deliberately partial fixture now has its own explicit dataset identity (`GRCh38-osteosarc-six-transcript-subset`) instead of claiming the full reference dataset's identity. Its coordinate assembly is still GRCh38 and provenance remains pinned. This isolates the pilot but does not fix Varcode's general cache bug or mutate its global cache. The upstream fix belongs here.

iskandr:
## Summary

Fixes #402. Patch release **8.0.3**, based on released 8.0.2 and independent
of the draft 9.0.0 cleanup in #439.

- Remove the assembly-name global contig cache. Validate using the actual
  Genome's `contigs()`; pyensembl already memoizes the database query per dataset.
- Preserve invalid-contig errors and normalization. Suppressed initial lookup
  errors now log their cause and return an empty `EffectCollection` with selected
  annotator provenance, not a plain list. No transcript is fabricated on failure.
- Add 34 regression cases using real, tiny, offline GTF-backed Genome datasets:
  both loading orders; identical names/versions but differing annotation paths;
  differing annotation versions; native and wrapped genomes; all four public
  locus accessors; repeated valid queries; invalid-contig rejection; and error
  collection type/provenance/cloning.
- Reproduce the downstream subset-then-full-GRCh38 TP53 coding deletion using
  cached Ensembl 81, verifying exact p.E2del protein reconstruction.

## Verification

- Before the fix: subset-first rejects valid chr17; an isolated full-first run
  incorrectly accepts chr17 in the subset; suppressed errors return a plain list.
- Targeted regression tests: **34 passed**.
- `./lint.sh`: passed.
- Full `TEST_SH_MAX=4 ./test.sh`: **1,406 passed**, no skips.
- GitHub CI: all six test jobs (push/PR, Python 3.9/3.10/3.11) and docs passed.

No biological prediction rules, public API removals, or Isovar fixture changes.
Release-script limitations are already tracked in #414.


iskandr:
Released as varcode 8.0.3: merged at 4e3973dcf9eeba7af24ffe0d24e5b5754d971856, all CI green, clean-main deploy passed lint/full suite, and wheel/sdist hashes match PyPI. A fresh public-PyPI installation passes all 34 new contig regressions, including exact TP53 p.E2del reconstruction, and both installed CLI help commands work with pkg_resources absent. Tag v8.0.3 is pushed. This also addresses the upstream return-type cause described in openvax/isovar#257.

Issue #261, opened 2026-04-11
Load structural variant sequences from Exacto


iskandr:
## Summary

Integrate varcode with [Exacto](https://github.com/pirl-unc/exacto) to support loading and representing structural variant (SV) sequences and their protein-level consequences. This requires expanding varcode beyond its current point-variant model (SNVs, indels) to handle multi-locus rearrangements, accept externally computed annotations, and incorporate RNA-level evidence for determining variant effects.

## Motivation

Varcode currently predicts variant effects from genomic coordinates alone, which works well for point variants but falls short for structural variants:

- **SVs have complex genomic signatures** — translocations, inversions, and duplications involve two or more breakpoints, potentially on different chromosomes, and don't fit the single-locus `(contig, start, ref, alt)` model
- **Protein consequences often can't be predicted from DNA alone** — a genomic translocation may or may not produce a fusion protein; this is only knowable from RNA evidence showing the chimeric transcript is actually expressed and translated
- **One DNA event can produce multiple spliced isoforms** — a translocation into a non-coding region might splice multiple ways, each producing a different protein. The reference sequences that matter may be intronic or intergenic, only becoming exonic in the novel rearrangement context
- **Exacto solves this** — it calls DNA and RNA variants from long-read data, integrates them, and translates full-length transcript structures into primary protein structures with complete provenance

## Key design principles

These principles should guide all sub-issues:

1. **DNA predicts possibilities, RNA resolves them.** A single DNA variant (especially an SV) can produce multiple transcript isoforms via alternative splicing. Without RNA data, the honest output is a *set of plausible outcomes*, not a single deterministic prediction. RNA evidence constrains this set to what was actually observed.

2. **This applies retroactively to existing splice handling.** Varcode today returns a single `SpliceDonor` or `ExonicSpliceSite` effect, but doesn't model what the resulting protein looks like. These effects should evolve toward a possibility-set model — and RNA evidence should be able to narrow that set.

3. **Reference sequence outside exonic regions matters.** SVs (and some splice variants) create novel transcripts that incorporate previously intronic or intergenic sequence. Effect prediction cannot be limited to annotated exon sequences — it needs access to genomic sequence around breakpoints and splice junctions.

4. **One-to-many is the natural relationship.** A single DNA variant maps to a set of possible/observed isoforms, each producing a potentially different protein. The data model (variants, effects, collections) must support this natively, not treat it as an edge case.

## Sub-issues

| Issue | Description |
|-------|-------------|
[...]

Issue #260, opened 2026-04-11
Add loader for Exacto output formats


iskandr:
## Background

[Exacto](https://github.com/pirl-unc/exacto) (EXacto Accurate Characterization of Transcriptomes and genOmes) is a long-read variant caller and integrator that produces DNA variants, RNA variants, integrated DNA/RNA variants, transcript structures, and translated primary structures. Varcode should be able to load these outputs to take advantage of Exacto's structural variant and transcript-level analysis.

## Scope

Implement loaders (analogous to `load_vcf()` and `load_maf()`) for the following Exacto TSV formats:

### 1. Annotated DNA variant calls
**Columns**: `variant_call_id, chromosome_1, position_1, chromosome_2, position_2, variant_type, variant_sequence, position_1_genic_region, position_1_annotation, position_2_genic_region, position_2_annotation`

- Maps to varcode `Variant` (for SNV/indel/MNV) or new SV types (#257) (for translocations, breakpoints, inversions, duplications)
- Annotations should be importable as external effects (#258)
- Note that `position_1_genic_region` can be non-exonic — SVs frequently involve breakpoints in intronic or intergenic regions, and the loader should not discard these

### 2. RNA variant calls
**Columns**: `variant_call_id, transcript_model_id, reference_transcript_ids, chromosome_1, position_1, strand_1, operation_1, chromosome_2, position_2, strand_2, operation_2, variant_size, variant_type, variant_sequence, ...`

- Variant types include: SNV, SKP (skipping), UTR, and RNA-level splice variants (fusion gene, circular RNA, cryptic exon, exon truncation, intron retention)
- These should map to RNA-informed effects (#259)
- **Multiple rows per DNA variant**: a single DNA event can produce multiple transcript models with different RNA variant calls, reflecting distinct spliced isoforms — the loader must preserve this one-to-many relationship, not collapse to a single "best" isoform

### 3. Integrated DNA/RNA variants
**Columns**: `transcript_model_id, reference_transcript_ids, rna_variant_call_id, dna_variant_call_id, distance, rna_variant_position, dna_variant_position`

- Links RNA observations to their underlying DNA variants
- Enables concordance tracking between DNA and RNA evidence
- This is where the one-DNA-event-to-many-isoforms mapping is made explicit — the loader should construct the full set of RNA-observed effects for each DNA variant

### 4. Transcript structures
[...]

Issue #259, opened 2026-04-11
Incorporate RNA-level evidence for variant effects


iskandr:
## Background

Varcode currently predicts variant effects purely from genomic (DNA) coordinates mapped against reference gene annotations. For many structural variants, the actual protein-level consequence **cannot be reliably determined from WGS alone**:

- A genomic translocation may or may not produce a fusion transcript — this depends on whether the rearrangement is transcribed and how the resulting RNA is spliced
- A fusion protein is obvious from RNA-seq even when the underlying genomic breakpoint is complex or ambiguous
- RNA-level events (cryptic exons, exon truncations, intron retention, circular RNAs) have direct protein consequences that are invisible to DNA-only analysis
- UTR extensions and alternative splicing around breakpoints can only be resolved with transcript-level data

## Core modeling problem: one variant, many possible isoforms

A single DNA event — especially an SV — can produce **multiple distinct spliced isoforms**, each with a different protein consequence. Consider a translocation whose breakpoint falls in a non-coding region: depending on which cryptic splice sites are used, the resulting transcript could include different amounts of intronic/intergenic sequence, skip different exons, or frame-shift at different points. The set of possible protein products is combinatorial.

**Without RNA evidence**, the best we can do is enumerate a **set of plausible splice outcomes** based on nearby canonical and non-canonical splice signals, scored by likelihood. This is a departure from varcode's current approach, which returns a single deterministic effect per transcript. But it's the honest representation — a `SpliceDonor` effect today doesn't tell you *what the resulting protein looks like*, just that splicing is probably disrupted.

**With RNA evidence**, the possibility set collapses to what was actually observed: which isoforms are expressed, at what levels, producing which proteins. RNA data constrains and resolves the ambiguity that DNA-only analysis cannot.

This principle — **DNA predicts possibilities, RNA resolves them** — applies retroactively to existing splice-site variant handling (see "Implications for existing code" below), not just to new SV types.

## Scope

Extend varcode's effect model to incorporate RNA-derived evidence, specifically:

### New RNA-informed effect types
- **FusionTranscript** / **FusionProtein**: chimeric transcript joining exons from two genes, with the resulting protein sequence
- **CircularRNA**: back-spliced transcript forming a circle
- **CrypticExon**: novel exon not present in reference annotations, often from sequence that was previously intronic or intergenic
- **ExonTruncation**: partial exon usage at or near a breakpoint
- **IntronRetention**: failure to splice out an intron, with protein consequence
[...]

PR #444, opened 2026-09-16
Preserve RNA deletion haplotypes and expose Exacto protein predictions

### LINKED ISSUE #441: RNAReadPhasingSource loses known deletion haplotypes when RNA aligners encode the omission as N
## Problem

`RNAReadPhasingSource._has_exact_deletion` / `_deletion_allele` requires the exact
CIGAR `D` interval. This loses RNA sequence support for a known DNA deletion when
a splice-aware aligner encodes the same omission as `N`, or an equivalent local
haplotype has another gap/mismatch decomposition. This also prevents discovering
the linked substitutions that change the coding sequence.

This is a sequence-evidence limitation, **not** a proposal to treat all splice
skips as genomic deletions. An RNA gap alone cannot establish the DNA mechanism.

## Reproducer (9.0.0)

```python
from types import SimpleNamespace
import pysam
from varcode.rna_read_phasing import RNAReadPhasingSource

source = object.__new__(RNAReadPhasingSource)
variant = SimpleNamespace(start=3, end=5, ref="AAA", alt="")
for gap in (2, 3):  # D versus N; identical query and reference endpoints
    read = pysam.AlignedSegment()
    read.reference_start = 0
    read.query_sequence = "ACGT"
    read.query_qualities = [40] * 4
    read.cigartuples = [(0, 2), (gap, 3), (0, 2)]
    print(gap, source._deletion_allele(read, variant))
# 2 alt
# 3 None
```

## Observed public-data example

GRCh38 MAP2: `chr2:209694773 delGCTACTGTGTGTTCAATAAGTACACAGT`, linked to
`chr2:209694769 C>A` and `chr2:209694770 T>G`.

The identical anchored local query sequence is represented as:

- `28D` in tumor WGS;
- `28N` in Illumina RNA;
- `22D + 2M + 6D` plus a mismatch in ONT RNA.

Exact joint local-sequence support with every retained query base >=Q20 is 24
T1 DNA reads, 16 T2 DNA reads, 4 T1 Illumina RNA reads, 2 T2 Illumina RNA reads,
and 1 T2 ONT read. Products are not pooled. The two adjacent substitutions change
the deletion-only protein prediction from `SQLEDLAHCHHL...` to `SQLEDRAHCHHL...`
on ENST00000360351.8 (residue 867 L>R). Both remain frameshifted.

The region also has a genuine annotated 19,449-base exon skip. That large skip
does not observe the local haplotype and must not count as deletion evidence.

## Suggested direction / regression coverage

An explicitly labeled anchored RNA haplotype-sequence comparison could retain
this evidence independently of CIGAR decomposition. Keep DNA-genotype support,
RNA-sequence support, and splice mechanism distinct; retain unknown outcomes
when anchors or sequence are missing. Test D/N-equivalent short omissions,
adjacent substitutions, split-gap representations, large annotated exon skips,
low-quality/missing-quality bases, and no transitive or mate-only joins.

This was found during the all-182 osteosarc.com phasing audit. No production
fix has been applied in this analysis task.

### LINKED ISSUE #443: RNAReadPhasingSource merges unrelated read groups sharing a QNAME
In Varcode 9.1.0, `RNAReadPhasingSource._fragment_alleles_for_pair` groups records only by `query_name`. Two unrelated libraries/read groups reusing a QNAME can therefore be merged into an apparent cis observation. `supports_variant` also deduplicates by QNAME alone and undercounts such fragments.

Minimal case: one alignment named `same_name`, RG=library1, covers only alternate site A; another alignment named `same_name`, RG=library2, covers only alternate site B. With min_alt_reads=1 the old grouping reports cis, despite no shared molecule.

Use `(RG, QNAME)` as the fragment key in support counting and pair grouping. Keep an empty RG sentinel for BAMs without that tag. Include a regression proving that distinct read groups neither create cis linkage nor collapse support counts.

Found while implementing the anchored MAP2 RNA-haplotype comparison in #441; the current follow-up branch includes the small fix and both regression tests.

iskandr:
## Summary

- Add explicit, reference-checked `RNAReadPhasingSource.register_haplotype` for known local substitution/deletion combinations. Match the complete anchored observed sequence on one alignment, independent of equivalent D/N/split-gap encoding. Apply quality/edge filters to every retained base; no mate assembly or transitive joins; unmatched contexts stay unknown.
- Import optional native Exacto primary-structure TSVs through the existing fusion loader. Preserve peptide IDs and multiple ORFs, validate base/codon/source coordinates, expose partial proteins with explicit completeness, and retain all per-base provenance. Predicted protein sequence is not experimental translation evidence.
- Fix QNAME collisions across read groups in support counts and pair grouping.
- Bump to 9.2.0; no new production dependencies or effect hierarchy.

Fixes #441 via an explicit known-haplotype mode; ordinary CIGAR behavior remains the default. Fixes #443.
Related to #259, #260, #261; these broader roadmaps remain open.

## Scientific checks

- MAP2 regression uses the observed local allele sequence with `28D`, `28N`, and `22D2M6D`, on both strands. Negative controls include the large annotated skip, altered/missing anchors, low/missing qualities, partial mates, and no transitive joins.
- Native primary-structure checks cover repeated amino acids on three codon rows, multiple peptide IDs, partial ends, splice-event rows, malformed coordinates/codons and translation past a stop.
- Three matching native Exacto example pairs validate codon-for-codon (393 aa each). Three additional public peptide groups lack their matching RNA-source groups; reported upstream as https://github.com/pirl-unc/exacto/issues/15, not bypassed by this loader.
- GABBR1 investigation remains separate local analysis: the early 384 coding nucleotides have one junction read, so reference-guided repair cannot be called an observed consensus. Raw and explicitly hypothetical protein translations are exported in the updated evidence bundle.

## Verification

- `./lint.sh`: passed.
- `./test.sh`: **1,506 passed**, no skips, 91% coverage.
- Raw-data analysis, FASTA exports and updated figures remain in the timestamped local output directory, not in the package.


iskandr:
Published as Varcode 9.2.0 on PyPI and tagged v9.2.0 at 5dab45af033aa1dcddb02feed0ce5939cb7a9a1d. Clean-main deploy reran lint and all 1,506 tests successfully; release CI is green. A fresh PyPI environment passed both CLI help checks and isolated native-primary/MAP2 API smoke checks. PyPI wheel/sdist hashes match the local build. External MAP2 validation reproduced the six previously audited anchored-context counts (1, 1, 4, 2, 24, 16); products remain separate. The expanded local evidence atlas passed 108 analysis regressions and contains explicitly labeled protein sequences.

Issue #418, opened 2026-09-15
SV effects report modifies_protein_sequence=False, so drop_silent_and_noncoding() drops fusions and large deletions


iskandr:
`StructuralVariantEffect` doesn't override `modifies_protein_sequence` or `modifies_coding_sequence`, so it inherits `False` from `MutationEffect` (`varcode/effects/effect_classes.py:104-109`). Every SV effect therefore claims to leave the protein unchanged, including `GeneFusion`, `LargeDeletion`, `LargeDuplication` and `Inversion`.

As a result, anything that filters on these flags drops SVs:

- **`EffectCollection.drop_silent_and_noncoding()`** filters on `modifies_protein_sequence` (`effect_collection.py:225`). `docs/effect_annotation.md:41` recommends it as the standard filter.
- **vaxrank's DNA fallback** keeps only effects with `e.modifies_protein_sequence` (`vaxrank/mutant_protein_fragment.py:377-383`), so an SV gets "No protein-modifying effects ... skipping".
- **isovar** exports `predicted_effect_modifies_protein_sequence` (`isovar_result.py:380-388`), and it will be `False` for SVs.

Repro, on main at 693f063, with the default annotator:

```python
g = cached_release(81)
cftr = g.transcript_by_id("ENST00000003084")
brca1 = g.transcript_by_id("ENST00000357654")
bnd = StructuralVariant(contig="7", start=cftr.start + 500, sv_type="BND",
                        mate_contig="17", mate_start=brca1.start + 1000, genome=g)
dele = StructuralVariant(contig="7", start=cftr.start + 100, end=cftr.start + 50_000,
                         sv_type="DEL", genome=g)
bnd.effects().drop_silent_and_noncoding()    # empty; GeneFusion dropped
dele.effects().drop_silent_and_noncoding()   # empty; LargeDeletion dropped
```

The point-variant path doesn't have this problem. On the same deletion, `ExonLoss` returns `True`, so passing `annotator="fast"` keeps it and the correct annotator drops it.

Proposed fix: return `True` from both properties for SV effects that touch coding exons or join two genes, and `False` when the SV only hits UTR or intron. Add a regression test that runs `drop_silent_and_noncoding()` on a fusion and on an exon-spanning deletion.

Issue #417, opened 2026-09-15
StructuralVariant.is_snv is True, and EffectCollection.to_dataframe() crashes on SVs


iskandr:
`StructuralVariant` inherits `is_snv`, `is_transition` and `is_transversion` from `Variant`, and those properties read its placeholder `ref="N"`, `alt="A"`. As a result:

- **`is_snv` is `True`** for every structural variant, including a 50 kb deletion.
- **`is_transition` and `is_transversion` raise** `ValueError: N is a non-standard nucleotide, neither purine or pyrimidine`.
- **`EffectCollection.to_dataframe()` crashes** on any collection that contains an SV, because `_DATAFRAME_COLUMNS` reads all three properties (`varcode/effects/effect_collection.py:339`, `:349-357`). `EffectCollection.to_csv()` goes through `to_dataframe()` (`:396`), so CSV export crashes too. This happens on the default annotation path, with no `annotator=` override.

Repro, on main at 693f063:

```python
from pyensembl import cached_release
from varcode import StructuralVariant, VariantCollection

g = cached_release(81)
cftr = g.transcript_by_id("ENST00000003084")
sv = StructuralVariant(contig="7", start=cftr.start + 100, end=cftr.start + 50_000,
                       sv_type="DEL", genome=g)
sv.is_snv                                          # True
VariantCollection([sv]).effects().to_dataframe()   # ValueError
```

`VariantCollection.to_dataframe()` works, because it doesn't read these properties.

Proposed fix: override `is_snv`, `is_indel`, `is_insertion`, `is_deletion`, `is_transition` and `is_transversion` on `StructuralVariant` to return `False`, and add a regression test that round-trips SV effects through `to_dataframe()` and `to_csv()`. `variant_kind()` in #416 already checks `is_structural` before `is_snv` to avoid this.

iskandr:
Update from main at 50eafde (8.0.0): `EffectCollection.to_dataframe()` and `to_csv()` no longer crash on SVs, but the placeholder alleles still reach the output. Both `VariantCollection.to_dataframe()` (so `varcode-genes`) and the effects CSV show them:
```python
vc = load_vcf("tests/data/osteosarc_esvee_somatic.vcf", genome="GRCh38", parse_structural_variants=True)
vc.effects().to_dataframe()
```
```
                         variant contig     start ref alt  is_snv  is_transversion  is_transition  gene_name  effect_type  effect
[...]

Issue #405, opened 2026-09-11
DUP/INV with one end in a transcript and no fusion partner get a misleading MutantTranscript


iskandr:
When a transcript contains only one end of a `DUP` or `INV`, the rearranged allele joins part of that transcript to whatever lies at the other end. If that's a sense-to-sense coding partner the annotator now reports a `GeneFusion`; otherwise it falls back to `LargeDuplication` / `Inversion`, whose `MutantTranscript` builders assume the whole event lies inside the transcript.

For a `DUP` that means the exons inside the span are duplicated *within the same transcript*. Before the orientation-aware fusion change, annotating the CPEB2::FAM193A junction from the osteosarc.com esvee calls (a 12 Mb `DUP`, chr4:2664477-15012987, Ensembl 95) on CPEB2 (ENST00000538197) gave:

```
LargeDuplication segments: pre_dup_including_body[0:2034], dup_body_copy[0:2034], post_dup[2034:6878]
```

That is, CPEB2 exons 1-3 repeated inside CPEB2, which is not the allele: the novel junction joins CPEB2's 5' part to sequence from the other end, and an intact CPEB2 copy remains. That junction now becomes a fusion, but any DUP or INV whose other end has no coding partner still gets this layout. `_build_inversion_mutant_transcript` has the same issue, marking the transcript's exons up to the breakpoint as inverted.

Possible fixes: for transcripts with exactly one end inside the span, build the `MutantTranscript` from the junction (retained part of this transcript plus unannotated sequence, like `TranslocationToIntergenic`), or report a truncation effect without a duplicated body.

PR #445, opened 2026-09-17
Fix symbolic SV affected spans without moving VCF junctions

### LINKED ISSUE #404: Symbolic <DEL>/<DUP> count the VCF padding base as part of the event
For symbolic structural-variant alleles, VCF `POS` is the padding base *before* the event: a `<DEL>` with `POS=P` and `END=E` removes bases `P+1..E`, and a `<DUP>` duplicates `P+1..E`.

`parse_symbolic_alt` sets `start=POS`, `end=END`, and the SV annotator then treats `[start, end]` as inclusive:

- `_cdna_ranges_kept_after_deletion` drops every base in `[min(start, end), max(start, end)]`, so the padding base is deleted too.
- `_cdna_ranges_within_sv` (used for DUP and INV bodies) clips exons to `[start, end]`, so the padding base is duplicated or inverted too.

The effect is one base, and it only matters when `POS` is the last base of an exon, where it removes or copies an exonic base the caller didn't. The breakend-junction model added alongside the osteosarc.com SV fixes uses the VCF convention (a deletion joins `start`, kept, to `end + 1`), so the two now differ by one base at that edge.

Fix: either shift `start` to `POS + 1` for symbolic DEL/DUP/INV at parse time, or have the range helpers exclude `start`. Either choice needs the existing CFTR deletion/duplication tests' expected offsets checked.

iskandr:
## Summary

Fixes #404. Version 9.2.1.

- Parsed symbolic DEL/DUP/INV/CNV (including CN0/CN3) retain POS/END for
  records and junctions but set affected_start=POS+1. Reuses the existing
  paired-breakend span contract; no new coordinate abstraction.
- Reject missing/non-increasing END instead of assigning the padding base
  a fabricated one-base event. This is a parser support boundary, not a
  claim that every valid VCF must contain END.
- Preserve direct-constructor defaults and paired-breakend behavior.
- Regress exact DEL/DUP and inversion-layout bases on both strands, all
  symbolic span types, JSON, and the retained CFTR exon-terminal anchor
  through the actual VCF loader and default annotator.

Primary reference: VCF 4.3 section 1.6.1
https://samtools.github.io/hts-specs/VCFv4.3.pdf

## Verification

- ./lint.sh
- 26 focused regression tests
- ./test.sh (full suite; log retained locally)

This is the first focused PR in the SV correctness series. Filtering #418,
SV metadata #417 and one-ended layouts #405 are not folded into this fix.


iskandr:
Released as Varcode 9.2.1: https://pypi.org/project/varcode/9.2.1/ . Green CI preceded merge; clean-main deploy passed lint and 1,532 tests. Published wheel and source hashes matched the build. Installed-wheel API regression and both CLI help checks passed.

Issue #438, opened 2026-09-15
Structural effect JSON serialization fails on missing primary_effects


iskandr:
## Bug

Serializing structural effects fails because `Serializable.to_dict()` infers
constructor argument names, including `primary_effects`, while
`StructuralVariantEffect` only stores `_primary_effects`.

Confirmed using the implementation from main commit
`9113af081b59978d9fa74832ad7f5d8053eff9ff`, before the structural-helper cleanup.
The failure also affects `EffectCollection.to_json()` containing these effects.

```python
from pyensembl import cached_release
from varcode import StructuralVariant

genome = cached_release(81)
variant = StructuralVariant(
    "7", 117531100, sv_type="DEL", end=117531200, genome=genome)
transcript = genome.transcript_by_id("ENST00000003084")
effect = variant.effect_on_transcript(transcript, annotator="fast")
effect.to_json()
# AttributeError: 'LargeDeletion' object has no attribute 'primary_effects'
```

The same missing-attribute error occurs in cleanup parity checks for DEL, DUP,
INV, CNV, INS, and BND returning TranslocationToIntergenic, with and without
`alt_assembly`.

## Fix requirements

- Define an explicit serialization/round-trip representation for structural
  effects, including primary, cryptic, splice and externally attached candidates,
  mutant transcript sequences/segments, and provenance.
- Do not simply expose `_primary_effects` as a property: its default is `(self,)`,
  so naive recursive serialization could cycle back into the parent effect.
- Cover direct effects and enclosing collections, including fusions, in tests.

Related: #417 covers SV dataframe export; this is the distinct JSON/object
serialization failure. Found while testing the default/structural cleanup;
not caused by that refactor and intentionally kept out of its scope.

iskandr:
Confirmed on published Varcode 9.3.7 while implementing openvax/vaxrank#482: serializing a `GeneFusion` now fails with `RecursionError` rather than the earlier missing-field error. Its `to_dict()` contains `candidates`, whose default primary points back to the same effect.

```python
from pyensembl import EnsemblRelease
from varcode import StructuralVariant, MutantTranscript
from varcode.effects import GeneFusion

g = EnsemblRelease(75)
t = g.transcript_by_id('ENST00000003084')
p = g.transcript_by_id('ENST00000269305')
v = StructuralVariant('7', 117171168, 'BND', mate_contig='17',
                      mate_start=7577121, genome=g)
e = GeneFusion(v, t, p, mutant_transcript=MutantTranscript(
[...]

PR #446, opened 2026-09-17
Preserve structural variant alleles and tabular metadata

### LINKED ISSUE #417: StructuralVariant.is_snv is True, and EffectCollection.to_dataframe() crashes on SVs

iskandr:
## Summary

Fixes #417. StructuralVariant now exposes its actual record REF/ALT and POS,
not the nucleotide placeholder used during initialization. Its six literal
small-edit flags are false; callers use `is_structural` / `sv_type`.

Mixed variant/effect tables retain symbolic or breakend alleles and add
`sv_type`, `end`, mate coordinates and affected-span coordinates. Point-only
and empty table schemas remain unchanged. Structural CSV summaries are
explicitly rejected on reimport instead of silently becoming point edits;
they are not a lossless RNA/effect archive (serialization work remains #438).

Version: 9.2.2.

## Verification

- `./lint.sh`: pass.
- Focused metadata, structural, CSV and partial-annotator tests: 153 passed.
- `TEST_SH_MAX=2 ./test.sh`: 1,578 passed; 91% coverage.
- Regressions cover A/C/G/T/N anchors, symbolic and breakend ALTs, JSON and
  pickle variant round-trips, mixed table exports, explicit CSV import errors,
  and unchanged point-only schemas.

Semantics follow VCF record REF/ALT fields, rather than interpreting symbolic
alleles as nucleotide substitutions:
https://samtools.github.io/hts-specs/VCFv4.3.pdf


iskandr:
Released as Varcode 9.2.2: https://pypi.org/project/varcode/9.2.2/ . Green CI preceded merge; clean-main deploy passed lint and 1,578 tests. Published artifact hashes and installed-wheel metadata/API plus both CLI help checks passed.

Issue #447, opened 2026-09-17
BND fallback advertises the whole reference transcript instead of the retained segment


iskandr:
## Reproduction

While fixing #405, I found that the separate BND fallback also advertises an
incorrect reference-segment layout. This is present on released 9.2.1.

```python
from pyensembl import cached_release
from varcode import StructuralVariant
g = cached_release(81)
t = g.transcript_by_id("ENST00000003084")
v = StructuralVariant("7", t.exons[0].start + 50, "BND",
    alt="N]22:15500000]", mate_contig="22", mate_start=15500000, genome=g)
e = v.effect_on_transcript(t)
print(type(e).__name__)
print([(s.label, s.start, s.end) for s in e.mutant_transcript.reference_segments])
```

Actual: `TranslocationToIntergenic`, `[('translocation_5p', 0, 6132)]`.
The left-retained breakpoint is at CFTR cDNA offset 50 (inclusive), so the
known 5-prime segment should be `[0:51]`, not the entire 6,132-base transcript.
The helper `_build_translocation_mutant_transcript` currently ignores both the
breakpoint and retained side and always emits the complete reference transcript.
The cDNA itself remains None, but downstream segment assembly sees false bases.

## Focused follow-up

Use retained-side orientation and the existing strand-aware `_cdna_cut` to
represent only the supported local segment (or leave the structure unresolved
when orientation is unknown). Keep the non-transcript partner and full coding
sequence unknown; preserve caller-supplied assemblies. Add all four
strand/retained-side combinations and exonic/intronic boundary tests.

This is separate from #405's local DUP/INV fallback and from #420's broader
protein-consequence classification. VCF retained-side semantics:
https://samtools.github.io/hts-specs/VCFv4.3.pdf section 5.4.

Issue #420, opened 2026-09-15
SV effect classes name the variant type, not the consequence; classify the mutant transcript instead


iskandr:
`LargeDeletion`, `LargeDuplication` and `Inversion` name the type of DNA event, not what happens to the transcript. The SV annotator emits them purely from geometry (`varcode/annotators/structural_variant.py:775-819`): if the SV is a DEL and overlaps an exon, it returns `LargeDeletion`. `INS` is reported as `LargeDuplication`, and `Inversion` doesn't even record which exons it hits. `ExonLoss` on the point-variant path is just as coarse: it fires whenever a variant spans more than one exon.

The consequence is the part that matters, and varcode can already compute it. `_build_deletion_mutant_transcript` builds the mutant cDNA, then leaves the protein to "downstream consumers". If that cDNA is translated and passed to the existing `classify_from_protein_diff`, you get real answers. On CFTR (`ENST00000003084`, Ensembl 81):

| Deletion | `fast` | `structural_variant` | classified from the mutant transcript |
|---|---|---|---|
| exon 1 (incl. ATG at 117480095) through intron 3 | `ExonLoss` | `LargeDeletion` | `StartLoss` |
| exon 5 + flanking intron (90 exonic bp) | `ExonLoss` | `LargeDeletion` | `Deletion`: in-frame, 1480→1450 aa |
| exons 5–6 (254 exonic bp) | `ExonLoss` | `LargeDeletion` | `FrameShift`, protein truncated to 171 aa |

Both annotators give all three rows the same label. For neoantigen prediction, the in-frame deletion and the frameshift are completely different cases.

Proposal:
- Have the SV annotator classify each `MutantTranscript` it builds, and return consequence effects (`FrameShift`, `Deletion`, `StartLoss`, `ExonLoss` when no protein can be derived). Put them in `candidates` when the breakpoints leave splicing ambiguous, for example when a breakpoint falls inside an exon.
- Keep `GeneFusion` and `TranslocationToIntergenic`, which do describe consequences.
- Keep `sv_type` on the variant and in candidate evidence, not in the effect class name.
- This also fixes #418, because the consequence classes already report `modifies_protein_sequence` correctly.

It's a breaking change to the effect classes (8.0). Nothing in vaxrank, topiary or isovar references `LargeDeletion` or `LargeDuplication` by name.

PR #448, opened 2026-09-17
Stop fabricating local DUP/INV transcripts for nonlocal junctions

### LINKED ISSUE #405: DUP/INV with one end in a transcript and no fusion partner get a misleading MutantTranscript

iskandr:
## Summary

Fixes #405. The shared local DUP/INV body builder now requires every junction
end to lie within the selected transcript. Previously, clipping the genomic
span invented a duplicated or inverted body inside that transcript, even when
the true adjacency led outside it. This also prevents whole enclosing spans
from being modeled as tandem copies of the full cDNA.

This is deliberately a small correction, not a new effect-class hierarchy:
the DNA-level LargeDuplication/Inversion remains, with `mutant_transcript=None`
when no sequence is established. That means unknown sequence, not no protein
change and not a proven truncation. Existing coding-partner fusion predictions,
supplied `alt_assembly` sequences, and wholly internal models are preserved.
Fusion secondary span candidates likewise lose only the unsupported local model.

Version: 9.2.3. Related broader classification/filtering remains #420/#418.
I also reproduced and filed the separate BND retained-segment bug as #447;
this PR does not reuse or extend that incorrect fallback.

## Verification

- `./lint.sh`: pass.
- 44 new focused cases: CFTR and reverse-strand BRCA1, either or both ends
  outside, symbolic/direct inputs, explicit assemblies, internal events, and
  shorter CBX5 isoforms whose gene nevertheless spans both endpoints.
- Existing osteosarcoma CPEB2::FAM193A and OTX1::KIF3C protein regressions
  still pass; assertions additionally exclude fabricated secondary models.
- `TEST_SH_MAX=2 ./test.sh`: 1,622 passed, 91% coverage.

Junction interpretation follows retained-side adjacency semantics in VCF 4.3
section 5.4: https://samtools.github.io/hts-specs/VCFv4.3.pdf


iskandr:
Released as Varcode 9.2.3: https://pypi.org/project/varcode/9.2.3/ . Green CI preceded merge; clean-main deploy passed lint and all 1,622 tests, with no skips and 91% coverage. Published artifact hashes matched; installed-wheel API and both CLI help checks passed. Separate follow-ups found during verification are #447 (BND retained segment) and #449 (the independent experimental transcript_model path clips nonlocal layouts and can classify a coding-span DUP as FivePrimeUTR). This PR fixes the default local structural builder, not that experimental pipeline.

Issue #449, opened 2026-09-17
Experimental transcript_model classifies a cross-boundary coding DUP as FivePrimeUTR


iskandr:
## Reproduction on the installed PyPI 9.2.2 wheel

```python
from pyensembl import cached_release
from varcode import StructuralVariant
g = cached_release(81)
t = g.transcript_by_id("ENST00000003084")
v = StructuralVariant("7", t.start - 100, "DUP", end=t.end - 50, genome=g)
e = v.effect_on_transcript(t, annotator="transcript_model")
print(type(e).__name__)
print([type(c.effect).__name__ for c in e.candidates])
print(e.modifies_protein_sequence)
```

Actual: `FivePrimeUTR`, `['FivePrimeUTR']`, `False`.

The event is chr7:117479863–117668615, crosses the transcript boundary and
includes the CFTR coding sequence. This does not prove a changed protein:
an intact copy can remain. But neither does it establish that the event is
solely a 5-prime UTR edit, or resolve the transcript created at its junction.

`GenomicLayout.apply_structural_variant` clamps the affected interval to its
available reference segments before applying DUP/INV. The transcript model
constructs a finite layout (50-base flanks), so the distal junction can be
outside that layout. This is a separate path from the default structural
builder corrected in #448 / #405. Current #448 code reproduces this same
experimental result; this is not a regression from that fix.

## Focused fix boundary

Reject or explicitly leave unresolved nonlocal DUP/INV edits whose junctions
cannot be represented by the available layout, rather than treating a clipped
body as the entire event. Preserve explicit assembled alleles and fully
represented local events. Add both-strand, left/right/enclosing and isoform
coverage regressions. Do not replace this with a blanket protein-modifying flag.

Related broader classification/filtering design: #420, #418, #423.

PR #450, opened 2026-09-17
Fix retained reference segments for BND fallback effects

### LINKED ISSUE #447: BND fallback advertises the whole reference transcript instead of the retained segment

iskandr:
## Summary

Fixes #447. The BND no-fusion fallback used to label the complete reference
transcript as its retained 5-prime segment, regardless of the breakpoint or
orientation. The CFTR reproduction advertised [0:6132] instead of [0:51].

Reuse the existing strand-aware cDNA cut and orientation helpers to retain
only the appropriate 5-prime prefix or 3-prime suffix, inclusive of an exonic
breakpoint. A small private parser helper also recognizes single-breakend
local orientation using the existing regex; paired-breakend parsing, event
typing and fusion selection are unchanged.

Unknown local orientation or a breakpoint outside the selected transcript
leaves the model unresolved. Supplied alt_assembly sequences still take
precedence. A reference-derived segment is marked
`sequence_status=retained_reference_fragment`; whole cDNA/protein remain None.
This is not a claim of a mature transcript, complete allele or protein truncation.

Version: 9.2.4. No new public API, effect class or dependency. The independent
experimental nonlocal DUP classification problem remains #449.

## Verification

- The CFTR regression failed before the change: actual [0:6132], expected [0:51].
- 81 new cases cover both transcript strands, all four paired-breakend ALTs,
  single breakends with inserted bases, exonic/interior/exon-edge/intronic
  coordinates, malformed or unknown orientation, mate-only orientation,
  unrelated transcripts and preservation of supplied assemblies.
- Focused tests including existing osteosarcoma fusion-protein regressions:
  142 passed.
- `./lint.sh`: pass.
- `TEST_SH_MAX=2 ./test.sh`: 1,703 passed, no skips, 91% coverage.

Primary coordinate semantics checked before implementation: VCF 4.3 sections
5.4 and 5.4.9, https://samtools.github.io/hts-specs/VCFv4.3.pdf .


iskandr:
Released as Varcode 9.2.4: https://pypi.org/project/varcode/9.2.4/ . Green Python 3.9/3.10/3.11 CI preceded merge. Clean-main deploy passed lint and all 1,703 tests (no skips, 91% coverage). Published wheel/source SHA256 values matched the build. The installed PyPI wheel passed both-strand retained-fragment checks through default and transcript_model BND paths, assembly/unknown-orientation checks, and both CLI help commands. The separate experimental nonlocal-DUP problem remains #449.

PR #461, opened 2026-09-18
Fix SV protein-change flags and preserve unresolved candidates (#418)

### LINKED ISSUE #418: SV effects report modifies_protein_sequence=False, so drop_silent_and_noncoding() drops fusions and large deletions

iskandr:
## Summary

Closes #418.

- Compute SV coding/protein-change flags from available proteins, mapped
  reference ORFs, and explicit RNA-import CDS boundaries. Pure deletions can
  establish CDS/start loss without a translated protein. Unmapped assemblies,
  partial BND fragments, and nonlocal DUP/INV models remain unresolved.
- Use `True` / `False` / `None` for changed / unchanged / unknown. Explicit
  `Unresolved` and unclassified cryptic-exon candidates also use `None`.
- Aggregate alternatives without recursion through self-containing SV sets:
  any changed candidate wins, otherwise unknown wins over unchanged. Preserve
  all candidates, producer order, stored sequences, and provenance.
- `drop_silent_and_noncoding(keep_unresolved=True)` retains unknowns by default.
  Pass `False` to require a positive protein-change prediction. Bump to 9.3.0
  because the explicit unknown status/default retention is visible API behavior.
- Document the API and execute its example in tests.

## Scientific boundaries

This evaluates existing transcript hypotheses; it does not create new ORFs,
reconstruct inversions/partial assemblies, rank hypotheses, or establish
expression/translation. A changed CDS can encode the same protein (including
synonymous and stop-retained changes). Reference:
[Ensembl consequence definitions](https://mart.ensembl.org/info/genome/variation/prediction/predicted_data.html).

Each comparison uses the transcript being annotated. For a fusion, the 5′
partner provides the initiation site/genetic code even when annotating its 3′
partner. Losing a deletion's annotated start is disruptive, without guessing
an alternative initiation site. Broader protein-effect classification remains
#420; #449's experimental clipping is not fixed here.

## Verification

- Before the implementation, 19 of the initial 25 regressions failed on 9.2.8.
- All seven audited CPEB2–FAM193A fusion isoforms now survive filtering.
- Added tests for both-strand coding/start/whole-transcript/UTR deletions,
  synonymous/stop-retained CDS changes, unchanged UTR fusions/duplications,
  mitochondrial translation, imported RNA ORFs, missing annotation/sequence,
  candidate alternatives/cycles, strict filtering, and provenance.
- `./lint.sh` passed; full `./test.sh`: **1,779 passed, no skips**, 91% coverage.
- `mkdocs build --strict` passed. GitHub CI will be checked before merge.

## Downstream follow-up

Confirmed and filed openvax/vaxrank#482: its DNA fallback collapses tied fusion
candidates before checking protein availability/change. Correct set-level flags
do not mean the first candidate's stored protein is changed. That selection
policy is outside this Varcode fix; callers needing all proteins should keep
the candidate set. Unknown status remains falsy in a raw Python truth test;
downstream callers must deliberately opt to retain it.


iskandr:
# Manual self-review — #418

No independent reviewer was used. Reviewed code, regressions, public behavior,
and the downstream callers after implementation.

- No blanket coding-overlap flag for DUP/INV/fusions. Nonlocal/partial models
  remain unknown; UTR-only predictions and identical proteins are not changed.
- CDS and protein are compared separately; tests cover synonymous and alternate
  stop codons. No unannotated ORF search, phase expansion, or candidate pruning.
- Reference initiation mapping requires all three bases in their original order;
  supports adjacent split segments and chooses the first retained copy.
- Both genomic strands exercised for deletions; translation uses the initiator's
  mitochondrial/nuclear table; either side of a fusion uses its own baseline.
- Empty protein is not missing protein. Missing CDS/start annotations and
  ambiguous sequences stay unknown rather than causing a false negative.
- Candidate graph traversal handles self references, back edges, explicit primary
  replacement, empty sets, and cycle-only sets. No candidate/provenance mutation
  and no caching that could hide later RNA candidates.
- Existing small-variant flags remain unchanged; explicit Unresolved and cryptic
  placeholders now honestly report None. Default filtering keeps unknowns;
  callers can request strict positive-only filtering. Documented minor bump.
- Strict docs build succeeds. A new test initially tried to patch a read-only
  transcript property; corrected the test to temporarily patch the descriptor.
- Residual downstream candidate-selection issue filed as openvax/vaxrank#482;
  the PR does not claim to fix all Vaxrank fusion-protein selection.

Remaining intentional scope: no new protein effect classes (#420), no fix to
experimental transcript clipping (#449), no SV serialization changes (#438).

Final local checks: lint passes; 1,779 tests pass with no skips (42.08 seconds,
91% coverage); strict documentation build passes. No remaining blocking finding
in the reviewed change. CI and published-wheel verification are still required.

iskandr:
Released as **[Varcode 9.3.0](https://pypi.org/project/varcode/9.3.0/)** after
merging the reviewed commit and running `./deploy.sh` from clean main.

- Lint and all **1,779 tests passed**, no skips; PR CI passed on Python 3.9,
  3.10, and 3.11. Strict docs build and deployed documentation passed.
- Verified the actual PyPI wheel in an isolated environment: both CLIs, all
  seven retained fusion candidates, changed/unchanged/unknown filtering,
  `keep_unresolved=False`, coding deletions, and ordinary L159M annotation.
- Three-valued flags and default unknown retention are documented in
  [SV filtering](https://openvax.github.io/varcode/structural_variants/#filtering-by-protein-change).
- Remaining consumer-selection bug is tracked separately in openvax/vaxrank#482.

Merge: fc82061dcaf9de5b20338fbeba0b298bc7e71f2c.

